PR #4, opened 2015-08-17
fix bucket listing permission

fiatala-avid:


cloudsploit-admin:
Hi, thanks for submitting a PR! Were you getting an error using the previous "ListBuckets" permission? It seems to work for me, so I'm trying to figure out the difference.

fiatala-avid:
Yep, s3 was returning a single UNKNOWN response. ListBuckets isn't in the official list of s3 permissions. I tested with the latest Amazon Linux image. Replacing with ListAllMyBuckets I get a partially working s3 evaluation. ACL and public writes are evaluated, but policy reads fail. I tried adding GetBucketPolicy  but that did not help. I may try a few more permissions next time I play with it. When I do I can paste the output.

cloudsploit-admin:
Hmm, odd. Thanks for trying to fix it too - I'll have a good place to start. I will try to get a fix up shortly for this once I'm able to see what permissions it actually needs.

cloudsploit-admin:
So it looks like having the following permissions allows the scan to run correctly:

```
"s3:GetBucketAcl",
"s3:ListAllMyBuckets"
```

Thanks for pointing out the original error with "ListBuckets." I'll update the README with the above permissions.

cloudsploit-admin:
Fixed in: 157d56f8e0

PR #5, opened 2015-08-17
Do not truncate ELB policies.

tpounds:
I'm not sure why this truncation logic was put into place (performance?). Since the scans depend on the official AWS SDK it appears the API pagination works as expected. Removing the logic from my fork returns results for all of our ELBs (> 100) as expected.

### DIFF
--- a/plugins/ec2/insecureCiphers.js
+++ b/plugins/ec2/insecureCiphers.js
@@ -153,16 +153,6 @@ module.exports = {
 					return rcb();
 				}
 
-				if (policies.length > 30) {
-					pluginInfo.tests.insecureCiphers.results.push({
-						status: 3,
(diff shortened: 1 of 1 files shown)


cloudsploit-admin:
Truncation was mainly for performance. Each of the checks is designed to be run from a Lambda function, so if it takes more than 60 seconds, it will fail. I found that doing 100 was a fair balance between  performance and getting adequate results, however it will obviously cause issues for users with > 100 ELBs. I will need to evaluate this change (and possibly combine it with async eachLimit) to perform simultaneous API calls.

tpounds:
Closing. It appears like someone merged this change out-of-band as part of fc4cc1c66bc9258ec289490a5d669ff4a8cfdc29.

PR #7, opened 2015-08-17
Use AWS credentials from environment by default.

tpounds:
Enables retrieving AWS credentials from the environment. Makes the Cloudsploit scans work as a pure [12-factor app](http://12factor.net/config). Really useful when externally injecting the config from configuration management and/or hosting inside of a container (Docker).

e.g.

```
docker run --env-file aws-creds.env cloudsploit > advisories.log
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,13 +1,13 @@
 var async = require('async');
 
-// var AWSConfig = {
-//     accessKeyId: '',
-//     secretAccessKey: '',
-//     sessionToken: '',
-//     region: 'us-east-1'
-// };
+var AWSConfig = {
+    accessKeyId:     process.env.AWS_ACCESS_KEY_ID,
+    secretAccessKey: process.env.AWS_SECRET_ACCESS_KEY,
+    sessionToken:    process.env.AWS_SESSION_TOKEN,
+    region:          process.env.AWS_DEFAULT_REGION
+};
 
-var AWSConfig = require(__dirname + '/../../cloudsploit-secure/scan-test-credentials.json');
+// var AWSConfig = require(__dirname + '/../../cloudsploit-secure/scan-test-credentials.json');
 
 var plugins = [
     'iam/rootAccountSecurity.js',


cloudsploit-admin:
Hi Trevor, thanks for the pull requests today. I'll look into the others shortly.

The changes you made above are actually already handled by the AWS SDK. I included the commented out settings for anyone who wanted to run the scans using the older key based method, but if none are provided (just leave them commented out), the SDK will check for environment variables and even IAM roles.

tpounds:
Thanks for taking a look. I get the following error with the commit 465a5c5 because it's looking for a json file which obviously doesn't exist.

```
module.js:338
    throw err;
          ^
Error: Cannot find module '../../cloudsploit-secure/scan-test-credentials.json'
    at Function.Module._resolveFilename (module.js:336:15)
    at Function.Module._load (module.js:278:25)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at Object.<anonymous> (/opt/cloudsploit/scans/index.js:10:17)
    at Module._compile (module.js:460:26)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Function.Module.runMain (module.js:501:10)
```

Commenting out the default config also doesn't work nor pickup my environment vars.

```
/opt/cloudsploit/scans/index.js:33
    plugin.run(AWSConfig, function(err, result){
               ^
ReferenceError: AWSConfig is not defined
    at /opt/cloudsploit/scans/index.js:33:16
    at iterate (/opt/cloudsploit/scans/node_modules/async/lib/async.js:146:13)
    at Object.async.eachSeries (/opt/cloudsploit/scans/node_modules/async/lib/async.js:162:9)
    at Object.<anonymous> (/opt/cloudsploit/scans/index.js:30:7)
    at Module._compile (module.js:460:26)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Function.Module.runMain (module.js:501:10)
    at startup (node.js:129:16)
```

IMHO, to make this work out of the box it should have a simple policy to fallback to retrieving credentials from multiple sources.

cloudsploit-admin:
The AWS Node SDK should take care of loading credentials from environment variables if they are set properly. This doesn't need to be done by the index.js file here because the SDK does it in the background.

The error above is caused because my credentials path is just a sample. You can change it to somewhere else if you're using it or just comment it out.

Here is the AWS guide on the algorithm the SDK follows to obtain credentials. Basically, it looks for hard-coded keys first, then a credentials path, then environment variables, then IAM roles.

http://docs.aws.amazon.com/AWSJavaScriptSDK/guide/node-configuring.html

I do agree with you that it should be able to be run from the first download, but there will need to be some configuration in order to get your credentials loaded. I can definitely do a better job of commenting that into the code, so I'll try to update that soon.

Let me know if you're able to get environment variables working.

tpounds:
I see what you are saying. The issue is that the index.js file is explicitly loading the [AWSConfig variable](https://github.com/cloudsploit/scans/blob/master/index.js#L33) prior to any of the SDK logic from loading. You can't use the index.js as an entrypoint and have it use the default SDK behavior. Does it make sense to remove the explicit config block entirely and instead rely on the SDK's default credential resolution behavior?

cloudsploit-admin:
Ah, yeah that makes sense. I didn't realize that if the roles / environment variables were being used, then AWSConfig would not actually be set.

Let me look into this more and see how to best approach it so it doesn't break anything else. What you suggested should work.

Thanks again!

maxbeatty:
+1 using environment variables for config. for non-docker users, [dotenv](https://github.com/motdotla/dotenv) is an option. with >iojs-v1.6.0, you can preload it so you don't need to include it in the source file

tpounds:
Closing this pull request in favor of #13.

PR #9, opened 2015-08-19
add comma to policy so that it is valid json

drewrothstein:


matthewdfuller:
Thanks for catching this!

PR #13, opened 2015-09-23
Use default AWS SDK credential resolution

tpounds:
This is an improvement to my original pull request (#7) to fetch credentials from the environment. I've updated index.js to fetch credentials using the method outlined in the [official AWS SDK documentation](http://docs.aws.amazon.com/AWSJavaScriptSDK/guide/node-configuring.html#Setting_AWS_Credentials). I think it's a minor improvement that makes the code a bit more portable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ npm install
 ```
 
 ## Setup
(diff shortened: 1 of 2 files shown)


cove:
+1

tpounds:
Discarding since this has been open for the better part of 4 years at this point.

PR #14, opened 2015-09-23
Remove debug log message.

tpounds:
This removes an extraneous debug message from the console log. AFAICT, the database plugin is the only one logging the region as it iterates. None of the others seem to do this.


matthewdfuller:
You're correct, none of the others do this. Was left behind by accident. Thanks for finding it!

PR #15, opened 2015-09-23
Remove API result limits

tpounds:
Removes a couple of API max item limits. AFAICT, there isn't any reason to limit the results.


matthewdfuller:
These limits are actually the default, so we'll still get back the same amount. However, I'll merge this in since AWS may increase the defaults in the future.

PR #16, opened 2015-09-26
allow the lambda function to use it's given execution role

kmcgrath:
If not crossing accounts the lambda function can be given a role upon creation, much like an ec2 profile, that can be used to grant read access to the current account.  In this case there is no need to pass credentials.


matthewdfuller:
This is fixed in the latest release along with a number of other Lambda updates, so I am closing this. Apologies for the delay, but thank you for contributing the PR.

PR #6, opened 2015-08-17
Do not truncate certificate metadata.

tpounds:
Similar to pull request #5 this removes the truncation logic for certificate metadata.

PR #17, opened 2015-09-28
Do not limit/truncate API results

tpounds:
Removes request limits/truncation across the remaining plugins. It is similar to pull requests #5, #6, and #15.

### DIFF
--- a/plugins/ec2/accountLimits.js
+++ b/plugins/ec2/accountLimits.js
@@ -189,18 +189,7 @@ module.exports = {
 						},
 
 						function(cb) {
-							// Query for the number of instances
-							if (limits['max-instances'] > 1000) {
-								pluginInfo.tests.instanceLimit.results.push({
-									status: 3,
-									message: 'Unable to query for more than 1000 instances',
-									region: region
-								});
-
-								return cb();
-							}
-
-							ec2.describeInstances({MaxResults:1000}, function(err, data){
+							ec2.describeInstances(function(err, data){
 								if (err || !data || !data.Reservations) {
 									pluginInfo.tests.instanceLimit.results.push({
 										status: 3,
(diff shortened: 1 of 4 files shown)


matthewdfuller:
I'm going to need to modify this one a bit because the limits checked for here (>1000 instances, 100 keys, etc.) are actually imposed by the AWS API. So theoretically, the code will need to check for the "NextToken" property of the response and then recursively call the same endpoint again until it is null. See: http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/EC2.html#describeInstances-property

tpounds:
I don't think this has to be modified as-is to be merged. It's only removing the limits and defers to the AWS defaults. That said, I do agree that every plugin should be updated to support proper pagination. I'm actually working on a series of commits to add support for that. I'll send a pull request for a single plugin so you can see the approach I am taking.

tpounds:
FWIW, we have a really large AWS account with lots of resources across multiple services. About half of the plugins are only returning partial results either due to truncation or lack of pagination support.

matthewdfuller:
Okay perfect, thanks so much for working on that. I agree - this would be preferable to adding an "unknown" result to some of them. My reasoning behind that was that at least you are *aware* things are being truncated. But with the added pagination support, we won't need this anymore.

tpounds:
I agree it's preferable to emit a warning than silently ignore subsequent results. I'll try to get the other changes in a pull request ASAP.

matthewdfuller:
@tpounds - Apologies for the (very late) ping on this PR, but we were performing an evaluation of code contributed to CloudSploit and realized we don't have a license agreement in place for you. Would you mind using the below link to review and digitally-sign CloudSploit's CLA? Thanks in advanced!

PR #18, opened 2015-09-28
Paginate API Calls

tpounds:
Currently a work in progress but I want to send a pull request to get early feedback. The idea is to change each plugin's API invocations to use the AWS SDK's [builtin pagination functionality](http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/Request.html#eachPage-property). It requires minor reorganization of each plugin's logic to support result aggregation.

### DIFF
--- a/plugins/ec2/accountLimits.js
+++ b/plugins/ec2/accountLimits.js
@@ -189,8 +189,27 @@ module.exports = {
 						},
 
 						function(cb) {
-							ec2.describeInstances(function(err, data){
-								if (err || !data || !data.Reservations) {
+							var reservations = 0;
+
+							ec2.describeInstances().eachPage(function(err, data){
+								if (!data) {
+									var returnMsgIl = {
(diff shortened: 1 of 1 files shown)


matthewdfuller:
I like this approach - it'll be easily reproducible across most of the APIs that support pagination. Thanks for putting it together. Would "eachPage()" be something we can write globally?

tpounds:
I'm not sure I quite understand what you mean by write "eachPage()" globally? Do you mean write the logic once and have it apply everywhere?

cove:
+1

matthewdfuller:
Apologies for the late followup; this functionality has been implemented: https://github.com/cloudsploit/scans/blob/d42d352961cab2e45d051afd44349244b0ba5ef6/collectors/aws/collector.js#L735-L757

Thank you for opening this issue with the code as well!

PR #28, opened 2016-05-26
Fix bug with Custom limits

cpollard0:
If a user has a custom limit in 1 region and not in another, async.each will reset the custom limit to the default. Change async.each to async.forEachSeries to fix.

### DIFF
--- a/plugins/ec2/instanceLimit.js
+++ b/plugins/ec2/instanceLimit.js
@@ -18,7 +18,7 @@ module.exports = {
 	run: function(AWSConfig, cache, callback) {
 		var results = [];
 
-		async.each(helpers.regions.ec2, function(region, rcb){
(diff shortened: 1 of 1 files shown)


cpollard0:
This change has to go to any plugin that has region specific limits. 

If you want to maintain the parallel processing of async, we can add a region to the limits object as an alternative. For example, limits[region]['max-instances'] solves the problem of async function calls overwriting previously set limits['max-instances'].

matthewdfuller:
Thank you for submitting this, @cpollard0 . Instead of removing the async each in favor of eachSeries (we want to keep the benefits of parallel processing while waiting for AWS API responses), I've moved the `limits` variable inside of the async function so that each region's execution is separate.

https://github.com/cloudsploit/scans/commit/420867b0ddfd44d29331ea0235a02b99896981fc

cpollard0:
Good solution!

PR #29, opened 2016-07-09
Check for open database ports

agnivesh:
Checks to see if the following databases have their ports open to the
Internet
- Cassandra
- CouchDB
- MongoDB
- Redis
- Riak

https://blog.shodan.io/its-still-the-data-stupid/


matthewdfuller:
Thanks for the PR @agnivesh! We'll look through and merge this soon. One thing we do need to do is confirm all existing services that would be checked in `ap-south-1` are supported. We'll update shortly.

PR #34, opened 2016-10-19
Update region.js with US East (Ohio) us-east-2

citizenjosh:
per http://docs.aws.amazon.com/general/latest/gr/rande.html#as_region


citizenjosh:
Update region.js with US East (Ohio) us-east-2 per http://docs.aws.amazon.com/general/latest/gr/rande.html#as_region

matthewdfuller:
Thanks @citizenjosh - I actually combined this into a single PR (https://github.com/cloudsploit/scans/pull/36) but these changes are now live.

PR #42, opened 2017-03-12
Small Updates to Multi-Subnet Plugin

matthewdfuller:
Fixing a few details in the description/info, as well as improving performance by using the local cache.

### DIFF
--- a/plugins/vpc/multipleSubnets.js
+++ b/plugins/vpc/multipleSubnets.js
@@ -3,17 +3,19 @@ var async = require('async');
 var helpers = require('../../helpers');
 
 module.exports = {
-	title: 'Multiple networks/subnets are used in a VPC',
+	title: 'VPC Multiple Subnets',
 	category: 'VPC',
-	description: 'Ensures that VPC has multiple networks which provides a layered architecture',
-	more_info: 'A single network within a VPC exposes of a risk of increasing the impact radius in case of a compromisation.',
+	description: 'Ensures that VPCs have multiple networks to provide a layered architecture',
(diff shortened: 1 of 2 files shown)


plugins/vpc/multipleSubnets.js
```diff
@@ -22,14 +24,10 @@ module.exports = {
 			LocalAWSConfig.region = region;
 			var ec2 = new AWS.EC2(LocalAWSConfig);
 
-			var params = {
-
-			};
-
-			ec2.describeVpcs(params, function(err, data){
-				if (includeSource) source[region] = {error: err, data: data};
+			helpers.cache(cache, ec2, 'describeVpcs', function(err, data) {
```
matthewdfuller: Using the local cache ensures we only call `describeVpcs` once per region across all plugins.

plugins/vpc/multipleSubnets.js
```diff
@@ -63,51 +70,42 @@ module.exports = {
...
-								message: 'Using ' + data.Subnets.length + ' subnets may not be sufficient for a multi-layered architecture',
+								message: 'Using ' + subnetData.Subnets.length + ' subnets may not be sufficient for a multi-layered architecture',
 								region: region,
 								resource: vpcId
 							});
 						} else {
 							results.push({
 								status: 2,
-								message: 'Only one subnet, one VPC is used.',
+								message: 'Only one subnet (' + subnetData.Subnets[0].SubnetId + ') in one VPC is used.',
```
matthewdfuller: Include the ID of the subnet in the message, but the VPC ID as the resource (allowing for easier suppressions)

PR #60, opened 2017-08-05
added CloudFront HTTPS Only Plugin

abhishek-jaiswal:


plugins/cloudfront/cloudfrontHttpsOnly.js
```diff
@@ -0,0 +1,50 @@
...
+        if (!listDistributions.data.length) {
+            helpers.addResult(results, 0, 'No CloudFront distributions found');
+            return callback(null, results, source);
+        }
+        // loop through Instances for every reservation
+        listDistributions.data.forEach(function(Distribution){
+            if (Distribution.DefaultCacheBehavior.ViewerProtocolPolicy == 'redirect-to-https'){
+                helpers.addResult(results, 0, 'CloudFront distributions ' + 
+                        'are configured to redirect non-HTTPS traffic to HTTPS', 'global',
+                        Distribution.Id)
```
matthewdfuller: Please use `distribution.ARN` instead of `distribution.Id`

plugins/cloudfront/cloudfrontHttpsOnly.js
```diff
@@ -0,0 +1,50 @@
...
+            }
+            else{
```
matthewdfuller: Can we add an `else if` here to check for `https` and set the result to 1 (warning) with the message `The CloudFront distribution is set to use HTTPS but not to redirect users accessing the endpoint over HTTP.`

plugins/cloudfront/cloudfrontHttpsOnly.js
```diff
@@ -0,0 +1,50 @@
...
+        }
+
+        if (!listDistributions.data.length) {
+            helpers.addResult(results, 0, 'No CloudFront distributions found');
+            return callback(null, results, source);
+        }
+        // loop through Instances for every reservation
+        listDistributions.data.forEach(function(Distribution){
+            if (Distribution.DefaultCacheBehavior.ViewerProtocolPolicy == 'redirect-to-https'){
+                helpers.addResult(results, 0, 'CloudFront distributions ' +
```
matthewdfuller: Can we make this singular? `CloudFront distribution is configured to redirect non-HTTPS traffic to HTTPS`.

plugins/cloudfront/cloudfrontHttpsOnly.js
```diff
@@ -0,0 +1,50 @@
...
+        listDistributions.data.forEach(function(Distribution){
+            if (Distribution.DefaultCacheBehavior.ViewerProtocolPolicy == 'redirect-to-https'){
+                helpers.addResult(results, 0, 'CloudFront distributions ' + 
+                        'are configured to redirect non-HTTPS traffic to HTTPS', 'global',
+                        Distribution.Id)
+            }
+            else{
+                helpers.addResult(results, 2, 'CloudFront distributions ' + 
+                        'are not configured HTTPS', 'global',
+                        Distribution.Id)
```
matthewdfuller: Change `Distribution.Id to Distribution.ARN`

abhishek-jaiswal:
squash the commit and added to single pull request

PR #73, opened 2017-11-13
Skipped RDS Read Replicas fix #72

### LINKED ISSUE #72: RDS Read Replicas Cannot Use Automatic Backups
According to the AWS docs:

> Yes, you can create a snapshot of a PostgreSQL Read Replica, but you cannot enable automatic backups

In the [rdsAutomatedBackups](https://github.com/cloudsploit/scans/blob/master/plugins/rds/rdsAutomatedBackups.js) plugin, we should skip such instances.

This can be done by checking for the presence of `ReadReplicaSourceDBInstanceIdentifier` in the RDS `describeDBInstances` call.

jkabhishek:
fix issue #72

### DIFF
--- a/plugins/rds/rdsAutomatedBackups.js
+++ b/plugins/rds/rdsAutomatedBackups.js
@@ -49,9 +49,9 @@ module.exports = {
 				// For resource, attempt to use the endpoint address (more specific) but fallback to the instance identifier
 				var db = describeDBInstances.data[i];
 				var dbResource = db.DBInstanceArn;
-				debugger;
+
 				// skip if it is read only replica Source Indentifier
-				if ('ReadReplicaSourceDBInstanceIdentifier' in db){
+				if (db.ReadReplicaSourceDBInstanceIdentifier){
 					continue;
 				}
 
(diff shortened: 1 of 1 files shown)


plugins/rds/rdsAutomatedBackups.js
```diff
@@ -49,6 +49,11 @@ module.exports = {
 				// For resource, attempt to use the endpoint address (more specific) but fallback to the instance identifier
 				var db = describeDBInstances.data[i];
 				var dbResource = db.DBInstanceArn;
+				debugger;
```
matthewdfuller: Can you remove the debugger?

plugins/rds/rdsAutomatedBackups.js
```diff
@@ -49,6 +49,11 @@ module.exports = {
...
+				// skip if it is read only replica Source Indentifier
+				if ('ReadReplicaSourceDBInstanceIdentifier' in db){
+					continue;
+				}
```
matthewdfuller: Can you change this to `if (db.ReadReplicaSourceDBInstanceIdentifier)`? The AWS API may set this property, even if it is `null`.

PR #80, opened 2017-11-27
added emptySecurityGroups plugin

jabhishek87:


plugins/ec2/emptySecurityGroups.js
```diff
@@ -0,0 +1,57 @@
...
+                // get instances for segurity groups
+                var describeGroupInstances = helpers.addSource(cache, source,
+                ['ec2', 'describeInstances', region]);
+                var resource = 'arn:aws:ec2:' + region + ':' + sg.OwnerId + ':security-group/' + sg.GroupId;
+
+                if (!describeGroupInstances.data.length) {
+                    helpers.addResult(results, 2, 'No Instances found ', region, resource);
+                } else {
+                    helpers.addResult(results, 0, 'Instances Found ', region, resource);
+                }
```
matthewdfuller: For this plugin, we should check that the rules for the group are not empty. You don't need to check ec2:describeInstances

PR #95, opened 2018-01-17
added Default VPC In Use plugin

jkabhishek:
Determines whether the default VPC is being used for launching EC2 instances.


matthewdfuller:
Closing in favor of https://github.com/cloudsploit/scans/pull/96 which resolved the conflicts + added another check

PR #105, opened 2018-04-13
Check conditional for SNS and allow multiple principals

maclennann:
Check the policy condition for an SNS topic (or similar). The AWS
docs have you set `Principal: *` with a SourceArn condition with
the ARN of the SNS topic.

Additionally, check all principals (both for global and cross-account)
if there are multiple AWS principals.

---

Some of our teams pipe messages from SNS to SQS. According to [the AWS docs](https://docs.aws.amazon.com/sns/latest/dg/SendMessageToSQS.html#SendMessageToSQS.sqs.permissions), the correct way of doing this is to set the Principal to * and create a condition with a SourceArn of the SNS topic. To be honest, I haven't tried setting the Principal to anything else (e.g. account). But I think if this is what the AWS docs recommend, this should be a supported configuration.

Additionally, some of our teams have multiple principals in their policy statements (e.g. two different roles in their account that can send messages) and this check was failing for them, since it was expecting a single string - even in `Principal.AWS`. I've updated it in a couple spots to accept an array of principals as well.

I tested this against a handful of our accounts and the results seem much more accurate. But let me know if I broke a use-case that I didn't think to test against. Happy to do some more tweaking.

### DIFF
--- a/plugins/sqs/sqsCrossAccount.js
+++ b/plugins/sqs/sqsCrossAccount.js
@@ -1,6 +1,40 @@
 var async = require('async');
 var helpers = require('../../helpers');
 
+function globalPrincipal(principal) {
+	if (typeof principal === 'string' && principal === '*') {
+		return true;
+	}
+
+	var awsPrincipals = principal.AWS;
+	if(!Array.isArray(awsPrincipals)) {
+		awsPrincipals = [awsPrincipals];
+	}
+
+	return awsPrincipals.some(p => (
+		p === '*' ||
+		p === 'arn:aws:iam::*'
+	));
+}
+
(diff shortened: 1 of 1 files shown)


matthewdfuller:
Thanks @maclennann - we'll run the tests we have on our end and merge this in once confirmed. We appreciate the effort to help improve these use cases!

jvanbrunschot-coolblue:
Any updates on this? We have the same issue with this in CloudSploit.

matthewdfuller:
Sorry for the delay @maclennann @jvanbrunschot-coolblue - we've deployed this in production, so let us know if you see any issues.

PR #107, opened 2018-04-16
Configures CloudFront HTTPS-Only to Be a PASS

### LINKED ISSUE #106: CloudFront HTTPS-Only Should Not Be a WARN Result
In [CloudFront scans][], a scan is configured to WARN when a CloudFront distribution is available only over HTTPS:

```javascript
else if (Distribution.DefaultCacheBehavior.ViewerProtocolPolicy == 'https-only'){
    helpers.addResult(results, 1, 'The CloudFront ' + 
        'distribution is set to use HTTPS but not to' + 
        'redirect users accessing the endpoint over HTTP', 'global',
        Distribution.ARN)
}
```

This contradicts the "More Info" of the scan:

> For maximum security, CloudFront distributions can be configured to only accept HTTPS connections or to redirect HTTP connections to HTTPS.

These are presented as equal options, and HTTPS-only is even mentioned first. Additionally, this contradicts the spirit of the compliance section:

> HIPAA requires all data to be transmitted over secure channels. CloudFront HTTPS redirection should be used to ensure site visitors are always connecting over a secure channel.

HTTPS-only also ensures site visitors are always connecting over a secure channel.

<sub>Also, there's a small typo in the scan that results in the WARN description rendering as "but not toredirect users".</sub>

[CloudFront scans]: https://github.com/cloudsploit/scans/blob/a7636237409f2966d427cd7946732386f11d35c9/plugins/cloudfront/cloudfrontHttpsOnly.js#L47-L51

chrisoverzero:
For the reasons stated in #106.

Closes #106.

### DIFF
--- a/plugins/cloudfront/cloudfrontHttpsOnly.js
+++ b/plugins/cloudfront/cloudfrontHttpsOnly.js
@@ -45,9 +45,8 @@ module.exports = {
                         Distribution.ARN)
             }
             else if (Distribution.DefaultCacheBehavior.ViewerProtocolPolicy == 'https-only'){
-                helpers.addResult(results, 1, 'The CloudFront ' + 
-                        'distribution is set to use HTTPS but not to' + 
-                        'redirect users accessing the endpoint over HTTP', 'global',
+                helpers.addResult(results, 0, 'The CloudFront ' + 
+                        'distribution is set to use HTTPS only.', 'global',
                         Distribution.ARN)
             }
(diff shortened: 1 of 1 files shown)


matthewdfuller:
Thanks @chrisoverzero - we'll merge and deploy this asap.

chrisoverzero:
Oh, hey, great. Thanks for the lightning turnaround.

PR #116, opened 2018-06-29
Report warning when an empty defeult VPC is found.

kallu:
Report a warning for empty default VPCs. Deleting VPC is easier if done before any resources are (accidentially) deployed.


matthewdfuller:
Thank you for the contribution @kallu and apologies for the (very!) slow response on this. We recently merged a separate PR for this check ([see here](https://github.com/cloudsploit/scans/pull/225)) which should provide similar functionality.

PR #117, opened 2018-07-02
Plugin ec2 max instances

Mav55:
Adding Plugin EC2 Max Instances, checks for two thresholds, a global one as an aggregate count of all regions as well as per region.


plugins/ec2/instanceMaxCount.js
```diff
@@ -0,0 +1,94 @@
...
+            regex: '^[1-2]{1}[0-9]{0,2}$',
+            default: 200
+        },
+		instance_count_region_threshold: {
+			name: 'Instance Count Region Threshold',
+			description: 'Checks for the number of running instances in each region and triggers a failing result if it exceeds the specified count',
+			regex: '^[1-2]{1}[0-9]{0,2}$',
+			default: 100
+		}
+	},
```
matthewdfuller: @Mav55 - for this, can we change it to be per-region based? In other words:
```
instance_count_region_threshold: {
		name: 'Instance Count Region Threshold',
		description: 'Checks for the number of running instances in each region and triggers a failing result if it exceeds the specified count. If a region is not set, the region count default will be used. Format: region-name:count. Example: us-east-1:10,us-west-2:20',
		regex: '^[a-z-1-3:,]{5}$',
		default: 'us-east-1:100,us-west-1:100' // ADD ALL REGIONS
	}
}
```
Then, in the `run` below, do a `split` on `,` and parse out the applicable names for each region. Fall back to 100 if the value is invalid.

Mav55:
OK, and leave the global setting as well?

matthewdfuller:
Right, so global can continue functioning as-is, just the region setting will allow the user to set each region separately, rather than only one for all regions.

Now that I look at it more, it may also be easier to create a separate setting like this, rather than trying to parse out the region from an error-prone string concatenation :
```
instance_count_region_threshold_ue1: {
		name: 'Instance Count Region Threshold: us-east-1',
		description: 'Checks for the number of running instances in us-east-1 and triggers a failing result if it exceeds the specified count.',
		regex: '^[1-2]{1}[0-9]{0,2}$',
		default: 100
	}
},
instance_count_region_threshold_uw2: {
		name: 'Instance Count Region Threshold: us-west-2',
		description: 'Checks for the number of running instances in us-west-2 and triggers a failing result if it exceeds the specified count.',
		regex: '^[1-2]{1}[0-9]{0,2}$',
		default: 100
	}
}
```
It might be a long list for all 13 (?) regions, but it provides maximum flexibility and error handling

Mav55:
Works for me. WIll get to it.

PR #118, opened 2018-07-02
Check if ACM certificates are successfully using DNS validation

kallu:
Loop through ACM certificates and report failed validations (FAIL) and when no DNS validation is used (WARN).


matthewdfuller:
Thank you for the contribution, @kallu - looks great!

PR #121, opened 2018-07-12
Plugin ec2 approved instance types

Mav55:
Added EC2 Instance Approved Types plugin, looks for instances launched that are NOT in the approved list, and returns a FAIL message.


Mav55:
Submitting an updated version that looks for disapproved types and reports on those types of instances which have been configured as disapproved by the account holder.

Mav55:
closing old pr will re-submit merging to new scanner code-base

PR #122, opened 2018-07-13
EBS snapshots

jcornell123:


matthewdfuller:
Thanks for submitting @jcornell123 - a few things:

1. It looks like the same `ebsEncryptedSnapshots.js` file was submitted at the root level and in the plugins directory.
2. In order to use the `DescribeSnapshots` API response data, the call should be added to the collect.js file, similar [to this](https://github.com/cloudsploit/scans/blob/master/collect.js#L91-L93).
3. The new plugin should be added to the [exports.js](https://github.com/cloudsploit/scans/blob/master/exports.js) file.

jcornell123:
Thanks @matthewdfuller.  Made the updates mentioned.

PR #124, opened 2018-07-18
Region update

jcornell123:
Reordered the regions to match AWS list.
updated services that are available in all regions to reflect that.

### DIFF
--- a/helpers/regions.js
+++ b/helpers/regions.js
@@ -5,7 +5,7 @@ var regions = [
 	'us-east-2',		// Ohio
 	'us-west-1',		// Northern California
 	'us-west-2',		// Oregon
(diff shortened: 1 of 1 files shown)


helpers/regions.js
```diff
@@ -23,7 +25,7 @@ module.exports = {
 	iam: ['us-east-1'],
 	route53: ['us-east-1'],
 	route53domains: ['us-east-1'],
-	s3: ['us-east-1'],
```
matthewdfuller: This is an oddity of the AWS API, but we have to use us-east-1 to query all buckets, even though they're created in other regions.

matthewdfuller:
Overall, looks good - just change the S3 region back to us-east-1, and we can merge.

PR #129, opened 2018-07-27
Plugin: KinesisStreamsEncrypted

philcorn:

PR #131, opened 2018-07-30
Kinesis Streams Encryption Plugin

jcornell123:
Unzipped the files and added the content from #129. PR129 can be closed and replaced with #130 and #131.

### DIFF
--- a/collect.js
+++ b/collect.js
@@ -219,6 +219,11 @@ var calls = {
 			override: true
 		}
 	},
+	Kinesis: {
+		listStreams: {
+			property: 'StreamNames'
+		}
+	},
 	KMS: {
 		listKeys: {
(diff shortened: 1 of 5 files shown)


collect.js
```diff
@@ -406,6 +411,21 @@ var postcalls = [
...
+		Kinesis: {
+			describeStream: {
+				reliesOnService: 'kinesis',
+				reliesOnCall: 'listStreams',
+				override: true
+			}
+		},
+		Firehose: {
+			describeDeliveryStream: {
+				reliesOnService: 'firehose',
```
matthewdfuller: This call relies on the firehose call, but I don't see it added above.

matthewdfuller: Nevermind, I see it in another PR

PR #136, opened 2018-08-23
filtering out duplicates

### LINKED ISSUE #135: KMS Key Policy Scan double-counts duplicates
My account has a KMS key that the KMS Key Policy plugin reports as having 13 users/roles allowed to use it. They key in question seems to have 3 policies, one with one user, another with 6 users, and a third with the same 6 users but different actions/conditions. I'd add that up to 7 instead of 13.

The code [does not seem to deduplicate](https://github.com/cloudsploit/scans/blob/master/plugins/kms/kmsKeyPolicy.js#L88) for each policy.

zmarois:
Fixes #135

### DIFF
--- a/plugins/kms/kmsKeyPolicy.js
+++ b/plugins/kms/kmsKeyPolicy.js
@@ -85,7 +85,13 @@ module.exports = {
 
 					if (!Array.isArray(principal.AWS)) continue;
 
-					totalUsers = totalUsers.concat(principal.AWS);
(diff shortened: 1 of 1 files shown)


matthewdfuller:
@zmarois - Apologies for the (very late) ping on this PR, but we were performing an evaluation of code contributed to CloudSploit and realized we don't have a license agreement in place for you. Would you mind using the below link to review and digitally-sign CloudSploit's CLA? Thanks in advanced!

zmarois:
Done

PR #137, opened 2018-08-27
Rds encrypted tde support

jcornell123:
Will check for TDE encryption as well as RDS native encryption.

### DIFF
--- a/collectors/rds/describeOptionGroups.js
+++ b/collectors/rds/describeOptionGroups.js
@@ -1,6 +1,5 @@
 var AWS = require('aws-sdk');
 var async = require('async');
-const _ = require("lodash");
 
 module.exports = function(AWSConfig, collection, callback) {
     var rds = new AWS.RDS(AWSConfig);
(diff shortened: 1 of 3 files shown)


collectors/rds/describeOptionGroups.js
```diff
@@ -0,0 +1,29 @@
+var AWS = require('aws-sdk');
+var async = require('async');
+const _ = require("lodash");
```
matthewdfuller: This is a heavy dependency for the smaller functionality below, can we re-write that using more native JS?

philcorn: @matthewdfuller - What makes this a heavy dependency?  The use of lodash?  Can you please provide more details so that this gets reworked inline with what is expected?

matthewdfuller: @philcorn - we would need to import the entire lodash module for the project just to do `_.find` in this one script. It's nothing against lodash specifically, but mostly as an internal practice we try to minimize third-party libraries where possible in favor of native JavaScript/Node. It helps reduce security concerns that often arise from the web of dependencies that Node modules have.

AkhtarAmir:
We will take the logic to create a new PR

PR #138, opened 2018-09-06
Add check for open IAM roles

maclennann:
Check IAM roles for AssumeRolePolicies that allow any AWS users to
assume the role. Checks for principal `AWS: *` in trust policy
statements.

---

I was reading https://rhinosecuritylabs.com/aws/assume-worst-aws-assume-role-enumeration/ a few days ago. I hadn't considered that an IAM administrator might misconfigure an AssumeRolePolicy to allow any AWS user to assume a role, but it seems like a serious issue that CloudSploit should have a check for.

### DIFF
--- a/collect.js
+++ b/collect.js
@@ -223,6 +223,10 @@ var calls = {
 		listUsers: {
 			property: 'Users'
 		},
+		listRoles: {
+			property: 'Roles',
+			override: true
+		},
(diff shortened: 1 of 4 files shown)


jcornell123:
This is a good plugin.  What's the status on being added in?

PR #141, opened 2018-10-02
Hotfixes/cloudtrail enabled plugin false positive 411

Mav55:
Here is a better approach to resolving this issue:

1. Will check for --include-global-service-events configuration
2. Will check if the Trail where that is configured is actually running 

Will report both issues on separate log entries under Global.

### DIFF
--- a/plugins/cloudtrail/cloudtrailEnabled.js
+++ b/plugins/cloudtrail/cloudtrailEnabled.js
@@ -2,78 +2,77 @@ var async = require('async');
 var helpers = require('../../helpers');
 
 module.exports = {
-	title: 'CloudTrail Enabled',
-	category: 'CloudTrail',
-	description: 'Ensures CloudTrail is enabled for all regions within an account',
-	more_info: 'CloudTrail should be enabled for all regions in order to detect suspicious activity in regions that are not typically used.',
-	recommended_action: 'Enable CloudTrail for all regions and ensure that at least one region monitors global service events',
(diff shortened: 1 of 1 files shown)


Mav55:
Sample messages:
- Configured but logging not enabled
CloudTrail	CloudTrail Enabled	N/A	global		FAIL	CloudTrail is configured for Global Monitoring in the us-west-1 region but is not logging API calls
CloudTrail	CloudTrail Enabled	N/A	global		OK	CloudTrail is configured to monitor global services

- Not configured
CloudTrail	CloudTrail Enabled	N/A	global		FAIL	CloudTrail is not configured to monitor global services

plugins/cloudtrail/cloudtrailEnabled.js
```diff
@@ -2,77 +2,87 @@ var async = require('async');
...
+                    var getTrailStatus = helpers.addSource(cache, source,
+                        ['cloudtrail', 'getTrailStatus', region, trail.TrailARN]);
+
+                    if (getTrailStatus && getTrailStatus.data &&
+                        getTrailStatus.data.IsLogging) {
+
+                        helpers.addResult(results, 0, 'CloudTrail is enabled', region);
+
+                        if (trail.IncludeGlobalServiceEvents) {
+                            helpers.addResult(results, 0, 'CloudTrail is enabled to monitor global services', region);
```
matthewdfuller: Wouldn't this duplicate results?

If `IncludeGlobalServiceEvents` is true, then a result gets added here and then again in line 82.

PR #144, opened 2018-10-18
Removing duplicate check for Sagemaker notebook encryption.

kallu:
Notebook encryption is checked in notebookDataEncrypted.js.


matthewdfuller:
@kallu - thank you for the fix (and sorry for the late reply) - it looks like enough has changed since you opened this PR to cause a conflict, so I've made the equivalent change in [this commit](https://github.com/cloudsploit/scans/commit/6ad5a1ae39f6765f5c827a60b1ef40a6f908156a). I'll close this PR then.

PR #146, opened 2018-10-26
EC2 Snapshots: Fix bug on collecting EC2 Snapshots

nancytao:
When data is null (i.e., an error is returned), data does not have an property 'Snapshots'. This fix prevents a TypeError from throwing.


matthewdfuller:
@nancytao - thank you for the contribution, good catch!

matthewdfuller:
@nancytao - Apologies for the (very late) ping on this PR, but we were performing an evaluation of code contributed to CloudSploit and realized we don't have a license agreement in place for you. Would you mind using the below link to review and digitally-sign CloudSploit's CLA? Thanks in advanced!

PR #160, opened 2019-03-11
Support for CIS compliance on AWS, suppression rules, exit code for CI integration, and CSV output

garretfick:
These changes are things that we have done for ourselves to integrate these scans with other tooling that we have. I'm happy to break this pull request into the several commits that actually happened.

The main changes here are:
1. added mapping to AWS CIS compliance rules
2. set a non-zero exit code if any rule failed
3. generate CSV output
4. suppress rules
5. start of adding unit tests for the test (because we like automated tests)

All of the following was done to be 100% compatible with the existing command line API. That is, if you use the command line as it is before this pull request (and afterward), you will get the exact same result.

The rules were also added in a way that doesn't touch the existing code - it is off to the side so that it would be very easy to add far more mappings to compliance rules without everyone editing the same files.


garretfick:
This adds fixes for path problems with running on Linux.

garretfick:
@matthewdfuller I've updated this to resolve recent merge conflicts

plugins/aws/iam/rootAccountInUse.js
```diff
@@ -27,6 +27,10 @@ module.exports = {
 	},
 
 	run: function(cache, settings, callback) {
+		this._run(cache, settings, callback, new Date())
+	},
+
+	_run: function(cache, settings, callback, now) {
```
matthewdfuller: Is there a benefit to doing this vs just:
```
run: function(cache, settings, callback) {
    var date = new Date();
    .... // rest of the code as-is
},
```

plugins/aws/iam/rootAccountInUse.js
```diff
@@ -75,11 +79,11 @@ module.exports = {
 					helpers.addResult(results, 0, 'Root account has not been used', 'global', obj.arn);
 				} else {
 					var dateToCompare = helpers.functions.mostRecentDate(accessDates);
-					var resultCode = (helpers.functions.daysAgo(dateToCompare) < config.root_account_in_use_days) ? 2: 0;
+					var resultCode = (helpers.functions.daysBetween(dateToCompare, now) < config.root_account_in_use_days) ? 2: 0;
```
matthewdfuller: Functionally this looks similar, but why was this changed to `daysBetween`?

garretfick: For testing. The function "daysAgo" always gets the current date/time. In order to test this, I needed a way to inject the current time so that I could test different scenarios. This approach allowed me to inject `now` into function for testing.

plugins/aws/iam/rootAccountInUse.js
garretfick: Yes - testability (as below). I wanted to have tests that inject different values for now so that I could check different conditions.

garretfick: You can see this in use with `rootAccountInUse.spec.js` which supplies a number of different dates for `now` and then evaluates what the rule says.

garretfick:
@matthewdfuller  I've been doing more testing today and it seems there is a problem here. I recommend to not merge this until I fix the issue in a couple hours.

garretfick:
I have fixed the bad merge, however there is still an issue that I'm unable to resolve.

In oracle\index.js, there is an import on line 8:

```
var logger = require(__dirname + '/../../../../helpers/logger.js');
```

This logger does not exist in the repo, and therefore fails to load. I haven't tried to address this, but master in your repo is currently broken for anyone.

matthewdfuller:
Hi @garretfick - apologies, we fixed this on a separate branch but I've just merged it in [here](https://github.com/cloudsploit/scans/commit/f2597299a5ecdbd16cd5a7d86d15a8701bfc9730). Please rebase.

garretfick:
Will do on Monday.

garretfick:
Rebasing with the merge and all caused me lots of headache, so I did the following:

1. Took your latest commit as the base
2. Replayed all of my changes against that commit (fixing conflicts along the way)
3. Force pushed my changes or overwrite our master, which are now effectively re-based from yours

garretfick:
@matthewdfuller any ETA if/when this could be merged. My hope is to keep things the same as what you have, but I also need to move forward on my objectives and cannot let this block things much longer.

garretfick:
Thanks!

PR #169, opened 2019-04-08
Support JUnit XML as an output format for integration of results with other CI services

garretfick:
This merge request adds support JUnit XML as an output format. This enables integration with services that can consume JUnit XML as an output, such as a CI service.

This is based off of another commit that includes support for various outputs. Once that get's merged, this changes here should look much smaller as there are only a 3 relevant commits.


garretfick:
I'll rebase and create a new pull request.

PR #172, opened 2019-04-12
Support JUnit XML as an output format for integration of results with other CI services

garretfick:
This pull request adds support JUnit XML as an output format. This enables integration with services that can consume JUnit XML as an output, such as a CI service.

Adds new tests in order to prove this works. Adds information to the README to explain how to generate JUNIT output.

This commit was rebased in order to have a clean history and resolve merge conflicts out of band.

I have tested this against a system I setup for verification. I've confirmed the JUnit file is correctly formatted and being parsed as expected. A 3rd party application shows the following expected result from that test system (cropped to show nothing interesting other than that the results can be parsed).

image

It would be possible to add validation against a schema for improved unit tests for this commit, but I decided against that based on prior discussions about testing and test layout.

### DIFF
--- a/index.js
+++ b/index.js
@@ -182,11 +182,13 @@ for (p in plugins) {
 			var plugin = getMapValue(serviceProviderPlugins, spp);
 			// Skip GitHub plugins that do not match the run type
 			if (sp == 'github' && serviceProviderConfig.org && !plugin.org) continue;
-			if (compliance.includes(spp, plugin)) {
-				for (pac in plugin.apis) {
-					if (serviceProviderAPICalls.indexOf(plugin.apis[pac]) === -1) {
-						serviceProviderAPICalls.push(plugin.apis[pac]);
-					}
+			
+			// Skip if our compliance set says don't run the rule
+			if (!compliance.includes(spp, plugin)) continue;
+
+			for (pac in plugin.apis) {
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -182,11 +182,13 @@ for (p in plugins) {
...
 			// Skip GitHub plugins that do not match the run type
 			if (sp == 'github' && serviceProviderConfig.org && !plugin.org) continue;
-			if (compliance.includes(spp, plugin)) {
-				for (pac in plugin.apis) {
-					if (serviceProviderAPICalls.indexOf(plugin.apis[pac]) === -1) {
-						serviceProviderAPICalls.push(plugin.apis[pac]);
-					}
+			
+			// Skip if our compliance set says don't run the rule
+			if (!compliance.includes(spp, plugin)) continue;
```
garretfick: This was a merge conflict - this change reduces indentation and follows the pattern from above.

index.js
```diff
@@ -252,6 +254,7 @@ async.map(serviceProviders, function (serviceProviderObj, serviceProviderDone) {
 	});
 }, function (err, results) {
 	// console.log(JSON.stringify(collection, null, 2));
+	outputHandler.close()
```
garretfick: This was a bug from before - without explicitly closing, then file was not actually written.

PR #176, opened 2019-04-17
Enable publishing cloudsploit scans as an npm package

garretfick:
This change makes it possible to publish cloudsploit as an npm package. It exposes a command line tool `cloudsploit-scan` that can be run.

This adds a new model of running cloudsploit - you can still do as it was done before with index.js directly. However, this new model expects you to use environment variables.

Publishing was setup using Travis CI and is triggered based on a git tag. Publishing was also setup so that forks which also have Travis CI setup to run tests can publish to a scoped package. That way, CloudSploit can still own the root (unnamed) scope, but others can create their own scope, and do this in a straightforward way.

Directions on how to do this are included in `.travis.yml` in the repository, since that's where someone might look if they wanted to know how to configure travis.

For now, I haven't updated the readme since I'm not convinced this is ready for primetime. I have no reason for not believing in the change other than a healthy dose of experience.


garretfick:
@matthewdfuller - the next change. This enables publishing cloudsploit to the NPM package repository. 

The next one after this will be to create a docker image that contains cloudsploit installed from NPM and configured to run easily, which is what I need for this to work a bit nicer with our CI server.

That work is already done (https://github.com/garretfick/scans/compare/feature/DOCKERIMAGE), so I know this all works once completed.

matthewdfuller:
Thank you for setting this up @garretfick - we had a few changes we were trying to get in ahead of merging, but everything is good to go now.

PR #185, opened 2019-05-23
Create and publish a docker image as a part of the travis build. 

garretfick:
This allows you to use a container to run scans without setting up NPM on your machine.

As before, this allows you to configure how the docker image is created using environment variables defined in Travis CI so that it is straightforward to make it work for forks.

The caveat is that this approach only installs the latest from NPM and creating the NPM package is based on tags. If this is undesirable, it is straightforward to only create the image if there is a tag.

@matthewdfuller

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,39 +4,73 @@ node_js:
   
 install: npm install
 
-# Deployment of cloudsploit is setup to work nicely both for the master repo
-# any forks. That is, a fork can also publish to NPM to a custom scope. To do
-# this, you need to do the following in the Travis CI settings for your account
-#
-# 1. Add an environment variable with the name NPM_SCOPE and a value like
-#
-#    @yourname\/
-#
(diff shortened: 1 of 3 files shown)


garretfick:
You can see a use of this change in my docker hub at https://cloud.docker.com/repository/docker/garretfick/cloudsploit-scans

Note that running this is currently broken due to issue https://github.com/cloudsploit/scans/issues/187

garretfick:
@matthewdfuller This could now be merged. I have confirmed that his builds cleanly on my branch.

garretfick:
@matthewdfuller Anything I can do to help move this along? This would substantially help our testing efforts.

matthewdfuller:
Thanks @garretfick - we were looking into setting up a Docker registry to make this work, but I realized that the travis code doesn't actually require those configs, it only pushes if it finds them, making this a much less time-intensive fix for us. I've merged, so let me know if you see any issues.

PR #186, opened 2019-05-23
Fix missing dependencies and refactor to really prevent in the future

garretfick:
@matthewdfuller 

A recent commit broke dependencies - @octocat/app is now required, but it is not in package.json.

This change adds that dependency and moves some code to a function so that nearly everything is callable from a test. I verified that the test fails if the dependency is not defined, so I'm certain that this will substantially reduce the risk in the future.

The moved code has no functional changes - it is just moved so that I can call from a unit test. However, this change is useful for the future so that it's possible to write more comprehensive tests.


garretfick:
I suspect that travis is not running all tests, but I propose to address that after pull request https://github.com/cloudsploit/scans/pull/185 since that also modifies the travis integration

PR #189, opened 2019-05-27
Fix some issues with detecting metrics

garretfick:
Also adds tests based on data returned from the AWS CLI. From my analysis, I don't think these scans should have ever succeeded, so I don't think these can break any existing uses.


garretfick:
@matthewdfuller

PR #198, opened 2019-07-20
fix executable name

jeshan:
The executable name is `cloudsploitscan`, not `cloudsploit-scan`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,5 +23,5 @@ ENV PATH "$PATH:/var/scan/node_modules/.bin"
 # command line arguments to the run command to control how this executes.
 # Thus, you can use the parameters that you would normally give to index.js
 # when running in a container.
-ENTRYPOINT ["cloudsploit-scan"]
+ENTRYPOINT ["cloudsploitscan"]
 CMD []


jeshan:
Based on the output of `ls` in the `/var/scan` directory:
```
lrwxrwxrwx    1 root     root            23 Jul 20 14:48 cloudsploitscan -> ../cloudsploit/index.js
lrwxrwxrwx    1 root     root            26 Jul 20 14:48 csv-write -> ../csv-write-stream/cli.js
lrwxrwxrwx    1 root     root            16 Jul 20 14:48 ndjson -> ../ndjson/cli.js
lrwxrwxrwx    1 root     root            20 Jul 20 14:48 semver -> ../semver/bin/semver
lrwxrwxrwx    1 root     root            23 Jul 20 14:48 sshpk-conv -> ../sshpk/bin/sshpk-conv
lrwxrwxrwx    1 root     root            23 Jul 20 14:48 sshpk-sign -> ../sshpk/bin/sshpk-sign
lrwxrwxrwx    1 root     root            25 Jul 20 14:48 sshpk-verify -> ../sshpk/bin/sshpk-verify
lrwxrwxrwx    1 root     root            16 Jul 20 14:48 uuid -> ../uuid/bin/uuid
lrwxrwxrwx    1 root     root            18 Jul 20 14:48 which -> ../which/bin/which
```

garretfick:
It would be create if you could make the build fail if this is not set correctly.

PR #203, opened 2019-09-10
Add a commandline arg to ignore results with an OK status

chrisfowles:
When running as a commandline tool (e.g. to verify infrastructure as part of development workflow) the number of OK results returned can make it difficult to filter through the noise.

This PR allows you to pass an argument `--ignore-ok` which will filter the output results to only statuses > 0.


chrisfowles:
Note I haven't added any tests around this behaviour as I couldn't see an easy way to implement testing around this area of the codebase.

If anyone has any pointers here; happy to add it in.

PR #206, opened 2019-10-22
Update lambda runtime eol

jcornell123:
Updated EOL for runtimes per https://docs.aws.amazon.com/lambda/latest/dg/runtime-support-policy.html


plugins/aws/lambda/lambdaOldRuntimes.js
```diff
@@ -18,8 +18,16 @@ module.exports = {
...
+            { 'id':'nodejs4.3-edge', 'name': 'Node.js 4.3 Edge', 'endOfLifeDate': '2018-04-30' },
+            { 'id':'nodejs6.10', 'name': 'Node.js 6.10', 'endOfLifeDate': '2019-04-30' },
+            { 'id':'nodejs8.10', 'name': 'Node.js 8.10', 'endOfLifeDate': '2019-12-31' },
+            { 'id':'dotnetcore2.0', 'name': '.Net Core 2.0', 'endOfLifeDate': '2019-04-30' },
+            { 'id':'dotnetcore1.0', 'name': '.Net Core 1.0', 'endOfLifeDate': '2019-06-27' },
+            { 'id':'python2.7', 'name': 'Python 2.7', 'endOfLifeDate': '2020-01-01' },
+            { 'id':'python3.5', 'name': 'Python 3.5', 'endOfLifeDate': '2020-09-13' },
+            { 'id':'python3.6', 'name': 'Python 3.6', 'endOfLifeDate': '2021-12-23' },
+            { 'id':'python3.7', 'name': 'Python 3.7', 'endOfLifeDate': '2023-06-27' },
+            { 'id':'python3.8', 'name': 'Python 3.8', 'endOfLifeDate': '2024-10-01' },
```
matthewdfuller: Hey @jcornell123 - these are accurate, but I realized we're not currently doing a date check further down in the code, only looking to see if the property exists in this map. Could you either comment out these lines (everything with an EOL > today) or update the below code to ensure future EOLs are skipped?

PR #208, opened 2019-11-07
Provide a Framework for running the scanning engine via AWS Lambda

jchrisfarris:
This PR provides a way to run the engine via AWS Lambda. A new file lambda_index.js is provided which contains the lambda_handler. A SAM CloudFormation Template and Makefile exist for deployment. A new method of delivering output as a json into AWS S3 is included. 

This work was done by Trek10 and sponsored by WarnerMedia for release back to the cloud security community.


matthewdfuller:
Thank you for the PR and contribution! Given the size, it may take us a bit to go through and test everything, but we will follow up if there are any questions.

Issue #210, opened 2019-11-18
Elasticsearch Logging Scan Produces False Negatives


chrisoverzero:
The test for Elasticsearch logging provides a different shape of data than the API actually provides. Here is a sample from one of the tests:

```javascript
{
  DomainStatus: {
    DomainName: 'mydomain',
    ARN: 'arn:1234',
    LogPublishingOptions: {
      Enabled: true,
      CloudWatchLogsLogGroupArn: 'arn:1234'
    }
  }
}
```

But the log publishing options are distinguished by kind of log, like this (anonymized from "Live Run" Raw Response):

```json
"LogPublishingOptions": {
  "ES_APPLICATION_LOGS": {
    "CloudWatchLogsLogGroupArn": "arn:aws:logs:<REGION>:<ACCOUNT>:log-group:<LOG_GROUP>:*",
    "Enabled": true
  },
  "INDEX_SLOW_LOGS": {
    "CloudWatchLogsLogGroupArn": "arn:aws:logs:<REGION>:<ACCOUNT>:log-group:<LOG_GROUP>:*",
    "Enabled": true
  },
  "SEARCH_SLOW_LOGS": {
    "CloudWatchLogsLogGroupArn": "arn:aws:logs:<REGION>:<ACCOUNT>:log-group:<LOG_GROUP>:*",
    "Enabled": true
  }
}
```

I expect this is causing every logging scan to fail for every user of CloudSploit with an ES resource.

chrisoverzero:
Can someone take a look at this and the associated PR? We have (and presumably everyone has?) had false failures for months because of this bug.

PR #212, opened 2019-11-18
Corrects Expected Shape of ES Domain Description

chrisoverzero:
The sources of logs are keys of the `LogPublishingOptions` object, and
they can be en- or disabled independently.

Fixes: #210


matthewdfuller:
Hey @chrisoverzero - thank you for these fixes and apologies for the slow response. We had a separate commit awhile back that made similar fixes ([see here](https://github.com/cloudsploit/scans/commit/fcfb6e0c51e9ef068e3af93e062ae1fa0753e76f#diff-98d6c347d3990ff770a1590953341438)) that should resolve these issues. Please let me know if that did the trick.

chrisoverzero:
Yes, that did do it. Thanks so much.

PR #219, opened 2020-01-09
Add JSON output option

joshlarsen:
Added option for JSON output and corresponding tests.


matthewdfuller:
Thank you for the contribution @joshlarsen!

PR #227, opened 2020-01-13
Feature/plugin/ebs volume private

jchrisfarris:
This PR Implements a check to ensure no EBS volumes are public. This check leverages Trusted Advisor, which requires Business or Enterprise support for API access. 

This plugin was developed by Trek10 under contract to WarnerMedia for release back into the main CloudSploit Scanning Engine. WarnerMedia expressly authorizes their contribution.


matthewdfuller:
Thank you for the contribution @jchrisfarris - we're going to merge this one as-is for now, but note that in the future it may be replaced by a plugin that leverages the EBS APIs with ARNs as the resource ID rather than relying on Trusted Advisor (most likely there will be duplicate plugins with an opt-in/out for this).

jchrisfarris:
That makes sense and is more broadly applicable. Trusted Advisor is only
available with Business/Enterprise support plans.

PR #229, opened 2020-02-04
Fix --json flag for JSON format

ssteo:
Fix for https://github.com/cloudsploit/scans/issues/228


matthewdfuller:
Thanks for the fix, @ssteo !

PR #233, opened 2020-03-03
Add --collection option that enables to save collected data

Pr0teus:
This pull  request creates a new option `--collection` that enables to save the data collected throught API requests and save in a json. It's useful to summary all the data collected and debug purposes.

How to use:
`node index.js --collection=./all-my-data.json`


matthewdfuller:
Thank you for the contribution @Pr0teus

PR #235, opened 2020-03-24
Update privateEndpoint.js

jeffthorne:
Error in recommended action text which referred to EKS clusters. Changing to GKE.


matthewdfuller:
Thank you for the contribution @jeffthorne

PR #267, opened 2020-08-10
Added plugin to mark lunch wizard security groups

AkhtarAmir:


plugins/aws/ec2/launchWizardSecurityGroups.js
```diff
@@ -0,0 +1,56 @@
...
+                helpers.addResult(results, 0, 'No security groups present', region);
+                return rcb();
+            }
+
+            for (var s in describeSecurityGroups.data) {
+                var sg = describeSecurityGroups.data[s];
+                // arn:aws:ec2:region:account-id:security-group/security-group-id
+                var resource = 'arn:aws:ec2:' + region + ':' + sg.OwnerId + ':security-group/' + sg.GroupId;
+
+                if (sg.GroupName.startsWith('launch-wizard')) {
```
matthewdfuller, line 40: Can we add a small check in here to lower-case this string first? Just in case they launch as `LAUNCH-WIZARD`

AkhtarAmir: Yes. let me add that

AkhtarAmir: I've made the changes. Kindly review it

PR #268, opened 2020-08-12
Added plugin for vpc endpoint acceptance for EC2

AkhtarAmir:


collectors/aws/collector.js
```diff
@@ -230,6 +230,13 @@ var calls = {
                 ]
             }
         },
+        describeVpcEndpointServices: {
+            property: 'ServiceDetails',
+            paginate: 'NextToken',
+            filters: {
+                serviceName: 'ec2'
+            }
```
matthewdfuller, line 238: These filters will need to be passed as params:
```suggestion
            params: {
                Filters: [
                    Name: 'serviceName',
                    Values: [
                        'ec2'
                    ]
                ]
            }
```

matthewdfuller: But I actually don't think you need this filter. The filter was for the name of the VPC endpoint service, not the name of the AWS service (i.e. `ec2`). You should be able to remove this param entirely.

plugins/aws/ec2/vpcEndpointAcceptance.js
```diff
@@ -0,0 +1,54 @@
...
+
+        async.each(regions.ec2, function(region, rcb){
+            var describeVpcEndpointServices = helpers.addSource(cache, source,
+                ['ec2', 'describeVpcEndpointServices', region]);
+
+            if (!describeVpcEndpointServices) return rcb();
+
+            if (describeVpcEndpointServices.err || !describeVpcEndpointServices.data) {
+                helpers.addResult(results, 3,
+                    'Unable to query for vpc endpoint services: ' + helpers.addError(describeVpcEndpointServices), region);
```
matthewdfuller, line 26:
```suggestion
                    'Unable to query for VPC endpoint services: ' + helpers.addError(describeVpcEndpointServices), region);
```

plugins/aws/ec2/vpcEndpointAcceptance.js
```diff
@@ -0,0 +1,54 @@
...
+                return rcb();
+            }
+
+            if (!describeVpcEndpointServices.data.length) {
+                helpers.addResult(results, 0, 'No vpc endpoint services present', region);
```
matthewdfuller, line 31:
```suggestion
                helpers.addResult(results, 0, 'No VPC endpoint services present', region);
```

plugins/aws/ec2/vpcEndpointAcceptance.js
```diff
@@ -0,0 +1,54 @@
...
+                return rcb();
+            }
+
+            for (var s in describeVpcEndpointServices.data) {
+                var service = describeVpcEndpointServices.data[s];
+                var resource = service.ServiceId;
```
matthewdfuller, line 37:
```suggestion
                var resource = service. ServiceName;
```
We want to use the ARN for the `resource` value for all results.

From the [docs](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/EC2.html#describeVpcEndpointServices-property):
> ServiceName — (String)
> The Amazon Resource Name (ARN) of the service.

plugins/aws/ec2/vpcEndpointAcceptance.js
```diff
@@ -0,0 +1,54 @@
...
+                if (service.AcceptanceRequired === false) {
```
matthewdfuller, line 38:
```suggestion
                if (!service.AcceptanceRequired) {
```

plugins/aws/ec2/vpcEndpointAcceptance.js
```diff
@@ -0,0 +1,54 @@
...
+                    helpers.addResult(results, 2,
+                        'Vpc endpoint service ' + (service.ServiceId) + ' does not require acceptance',
```
matthewdfuller, line 40:
```suggestion
                        'VPC endpoint service ' + (service.ServiceId) + ' does not require acceptance by the service owner',
```

plugins/aws/ec2/vpcEndpointAcceptance.js
```diff
@@ -0,0 +1,54 @@
...
+                        region, resource);
+                } else {
+                    helpers.addResult(results, 0,
+                        'Vpc endpoint service ' + (service.ServiceId) + ' requires acceptance',
```
matthewdfuller, line 44:
```suggestion
                        'VPC endpoint service ' + (service.ServiceId) + ' requires acceptance by the service owner',
```

plugins/aws/ec2/vpcEndpointAcceptance.js
matthewdfuller: Usually it's preferred to check for the passing condition first and then use the `else` for the failing condition. For example:
```
if (service.AcceptanceRequired) {
    // PASS
} else {
    // FAIL
}
```

plugins/aws/ec2/vpcEndpointAcceptance.js
```diff
@@ -0,0 +1,57 @@
...
+
+            if (!describeVpcEndpointServices.data.length) {
+                helpers.addResult(results, 0, 'No VPC endpoint services present', region);
+                return rcb();
+            }
+
+            for (var s in describeVpcEndpointServices.data) {
+                var service = describeVpcEndpointServices.data[s];
+
+                if (service.Owner === 'amazon') continue;
```
matthewdfuller, line 38: This `continue` is okay, but the issue is that if _all_ of the endpoints are owned by `amazon` then we will have no results because everything was skipped (every plugin should produce at least 1 result per region).

To fix this, you can either filter our the array before doing the loop (and check for array.length). Or, you can set a variable like:
```
var found = false;
```
and then, if the code makes it to line 39, `found = true`

and at the end, check `if (!found) helpers.addResult(results, 0, 'No user-owned endpoints found'...`

PR #272, opened 2020-08-14
Added plugin and spec file for launch wizard security groups | Removed duplicate test case from vpcEndpointAcceptance.spec.js

AkhtarAmir:


plugins/aws/cloudformation/plainTextParameters.js
```diff
@@ -0,0 +1,77 @@
...
+    more_info: 'CloudFormation supports the NoEcho property for sensitive values, which should be used to ensure secrets are not exposed in the CloudFormation UI and APIs.',
+    link: 'https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/parameters-section-structure.html',
+    recommended_action: 'Update the sensitive parameters to use the NoEcho property.',
+    apis: ['CloudFormation:describeStacks'],
+    settings: {
+        plainTextParameters: {
+            secretWords: [
+                'secret', 'password', 'privatekey'
+            ]
+        }
```
matthewdfuller, line 17: This is a good start, but see here for an example of the format:
https://github.com/cloudsploit/scans/blob/master/plugins/aws/acm/acmCertificateExpiry.js#L12-L18

For example, this section should be:
```suggestion
        plain_text_parameters: {
            name: "CloudFormation Plaintext Parameters",
            description: "A comma-delimited list of parameter strings that indicate a sensitive value",
            regex: "[a-zA-Z0-9,]",
            default: "secret,password,privatekey"
        }
```

Then later, you can create an array by doing a `split(',')` on the words

plugins/aws/cloudformation/plainTextParameters.js
```diff
@@ -0,0 +1,77 @@
...
+            }
+            
+            var parameterFound;
+            for (var s in describeStacks.data){
+                // arn:aws:cloudformation:region:account-id:stack/stack-name/stack-id
+                var stack = describeStacks.data[s];
+                var resource = stack.StackId;
+                parameterFound = false;
+
+                if(!stack.Parameters.length) {
```
giorod3, line 51: Add checks to make sure the key exists before referencing. Like this:
if(!stack.Parameters || !stack.Parameters.length) {

plugins/aws/cloudformation/plainTextParameters.js
```diff
@@ -0,0 +1,77 @@
...
+                    helpers.addResult(results, 0,
+                        'The template does not contain any potentially-sensitive parameters', region, resource);
+                    return rcb();
+                }
+
+                stack.Parameters.forEach(function(parameter){
+                    if(!parameterFound && secretWords.includes(parameter.ParameterKey.toLowerCase())) {
```
giorod3, line 58: Add checks to make sure the key exists before referencing. Like this:
if(!parameterFound && parameter.ParameterKey && secretWords.includes(parameter.ParameterKey.toLowerCase())) {

exports.js
```diff
@@ -14,6 +14,8 @@ module.exports = {
         'cloudfrontLoggingEnabled'      : require(__dirname + '/plugins/aws/cloudfront/cloudfrontLoggingEnabled.js'),
         'cloudfrontWafEnabled'          : require(__dirname + '/plugins/aws/cloudfront/cloudfrontWafEnabled.js'),
 
+        'plaintextParameters'           :  require(__dirname + '/plugins/aws/cloudformation/plaintextParameters.js'),
```
giorod3, line 17: the file is named "plainTextParameters" notice the uppercase T. This is also why the test is failing

plugins/aws/cloudformation/plainTextParameters.js
```diff
@@ -0,0 +1,77 @@
...
+                // arn:aws:cloudformation:region:account-id:stack/stack-name/stack-id
+                var stack = describeStacks.data[s];
+                var resource = stack.StackId;
+                parameterFound = false;
+
+                if(!stack.Parameters || !stack.Parameters.length) {
+                    helpers.addResult(results, 0,
+                        'Template does not contain any potentially-sensitive parameters', region, resource);
+                    return rcb();
+                }
```
giorod3, line 55: Do not add `return rcb();` in the for loop, we want to keep looping through describeStacks.data and `rcb();` should only be called when the function is completely finished for the entire **region**. 
Suggested change: 
`if(!stack.Parameters || !stack.Parameters.length) {
                    helpers.addResult(results, 0,
                        'Template does not contain any parameters', region, resource);
                        continue;
                }`

plugins/aws/ec2/launchWizardSecurityGroups.js
```diff
@@ -0,0 +1,55 @@
...
+            if (!describeSecurityGroups) return rcb();
+
+            if (describeSecurityGroups.err || !describeSecurityGroups.data) {
+                helpers.addResult(results, 3,
+                    'Unable to query for security groups: ' + helpers.addError(describeSecurityGroups), region);
+                return rcb();
+            }
+
+            if (!describeSecurityGroups.data.length) {
+                helpers.addResult(results, 0, 'No security groups present', region);
```
giorod3, line 31: Suggested change: 'No security groups found'

plugins/aws/ec2/launchWizardSecurityGroups.js
```diff
@@ -0,0 +1,55 @@
...
+                var sg = describeSecurityGroups.data[s];
+                var resource = sg.GroupId;
+
+                if (sg.GroupName && sg.GroupName.toLowerCase().startsWith('launch-wizard')) {
+                    helpers.addResult(results, 2,
+                    'Security Group ' + sg.GroupName + ' was launched using EC2 launch wizard',
+                    region, resource);
+                } else {
+                    helpers.addResult(results, 0,
+                        'Security Group ' + sg.GroupName + ' was not launched using EC2 launch wizard',
```
giorod3, line 45: `(sg.GroupName || '')` just incase there is no GroupName the result will show "Security Group  was not launched using EC2 launch wizard" instead of "Security Group undefined was not launched using EC2 launch wizard"

plugins/aws/cloudformation/plainTextParameters.js
```diff
@@ -0,0 +1,77 @@
...
+                helpers.addResult(results, 0, 'No stack description found', region);
+                return rcb();
+            }
+            
+            var parameterFound;
+            for (var s in describeStacks.data){
+                // arn:aws:cloudformation:region:account-id:stack/stack-name/stack-id
+                var stack = describeStacks.data[s];
+                var resource = stack.StackId;
+                parameterFound = false;
```
giorod3, line 49: `let parameterFound = false;`

plugins/aws/cloudformation/plainTextParameters.spec.js
```diff
@@ -0,0 +1,155 @@
+var expect = require('chai').expect;
+const plaintextParameters = require('./plaintextParameters');
```
giorod3, line 2: forgot this one. `./plainTextParameters`

plugins/aws/cloudformation/plainTextParameters.js
```diff
@@ -0,0 +1,76 @@
...
+                    helpers.addResult(results, 0,
+                        'Template does not contain any parameters', region, resource);
+                    continue;
+                }
+
+                stack.Parameters.forEach(function(parameter){
+                    if(!parameterFound && parameter.ParameterKey && secretWords.includes(parameter.ParameterKey.toLowerCase()) && !parameter.ParameterValue.match('^[*]+$')) {
+                        parameterFound = true;
+                        helpers.addResult(results, 1,
+                            'Template contains one of the following potentially-sensitive parameters: secret, key, password', region, resource);
```
matthewdfuller, line 60: These shouldn't be hard-coded because the user can override the `settings` section. Instead, you should keep an array of strings that it matches and then print them at the end. Something like:
```
var foundStrings = [];
if (...matches...) {
    foundStrings.push(ParameterKey);
}
```

plugins/aws/cloudformation/plainTextParameters.spec.js
```diff
@@ -0,0 +1,155 @@
...
+            StackName: 'TestStack',
+            Parameters: [
+                {
+                    ParameterKey: 'Secret',
+                    ParameterValue: 'bucketwithsecretparameter1'
+                },
+                {
+                    ParameterKey: 'Password',
+                    ParameterValue: 'bucketwithsecretparameter1'
+                }
```
matthewdfuller, line 16: We also need a test case with `***` to test that condition.

plugins/aws/cloudformation/plainTextParameters.js
```diff
@@ -0,0 +1,76 @@
...
+                }
+
+                stack.Parameters.forEach(function(parameter){
+                    if(parameter.ParameterKey && secretWords.includes(parameter.ParameterKey.toLowerCase()) && !parameter.ParameterValue.match('^[*]+$')) {
+                        foundStrings.push(parameter.ParameterKey);
+                    }
+                });
+
+                if(foundStrings && foundStrings.length) {
+                    helpers.addResult(results, 1,
```
matthewdfuller, line 62:
```suggestion
                    helpers.addResult(results, 2,
```

PR #278, opened 2020-08-22
Feature/sploit 681 | Added ElasticBeanstalk Managed Platform Updates plugin, spec files

AkhtarAmir:


plugins/aws/elasticbeanstalk/managedPlatformUpdates.js
```diff
@@ -0,0 +1,82 @@
...
+                }
+
+                for (var p in describeConfigurationSettings.data.ConfigurationSettings) {
+                    var param = describeConfigurationSettings.data.ConfigurationSettings[p];
+
+                    for (var s in param.OptionSettings) {
+                        var setting = param.OptionSettings[s];
+
+                        if (setting.Namespace === 'aws:elasticbeanstalk:managedactions' &&
+                            setting.OptionName === 'ManagedActionsEnabled') {
```
giorod3, line 62: check that `setting.Namespace` and `setting.OptionName` both exist before doing a comparison

plugins/aws/elasticbeanstalk/managedPlatformUpdates.js
```diff
@@ -0,0 +1,82 @@
...
+                            if (setting.Value === 'true') {
```
giorod3, line 63: check that `setting.Value` exists first

plugins/aws/elasticbeanstalk/managedPlatformUpdates.js
```diff
@@ -0,0 +1,82 @@
+var async = require('async');
+var helpers = require('../../../helpers/aws');
+
+module.exports = {
+    title: 'ElasticBeanstalk Auto Update',
```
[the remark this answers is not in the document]
AkhtarAmir: Actually there is nothing as Auto update in elastic beanstalk. I looked into the documentation after we had a discussion the week before previous, according to my understanding, we are looking for managed updates for environments.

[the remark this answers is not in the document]
giorod3: You can change the title name then to fit the name of the file.

AkhtarAmir:
Yeah thats what I was thinking that we should change the name of plugin

plugins/aws/elasticbeanstalk/managedPlatformUpdates.js
```diff
@@ -0,0 +1,82 @@
+var async = require('async');
+var helpers = require('../../../helpers/aws');
+
+module.exports = {
+    title: 'ElasticBeanstalk Managed Platform Updates',
+    category: 'ElasticBeanstalk',
+    description: 'Ensures ElasticBeanstalk applications are configured to managed updates.',
```
matthewdfuller, line 7:
```suggestion
    description: 'Ensures ElasticBeanstalk applications are configured to use managed updates.',
```

plugins/aws/elasticbeanstalk/managedPlatformUpdates.js
```diff
@@ -0,0 +1,82 @@
...
+    more_info: 'Environment for an application should be configured to allow platform managed updates.',
```
matthewdfuller, line 8:
```suggestion
    more_info: 'Environments for an application should be configured to allow platform managed updates.',
```

plugins/aws/elasticbeanstalk/managedPlatformUpdates.js
```diff
@@ -0,0 +1,82 @@
...
+                    'No application environments found', region);
+                return rcb();
+            }
+
+            async.each(describeEnvironments.data, function(environment, ecb){
+                var resource = environment.EnvironmentArn;
+                var describeConfigurationSettings = helpers.addSource(cache, source, ['elasticbeanstalk', 'describeConfigurationSettings', region, environment.EnvironmentName]);
+                if (!describeConfigurationSettings) return ecb();
+
+                if (describeConfigurationSettings.err ||
```
matthewdfuller, line 41: These should be combined as:
```suggestion
                if (!describeConfigurationSettings || describeConfigurationSettings.err ||
```

plugins/aws/elasticbeanstalk/managedPlatformUpdates.js
```diff
@@ -0,0 +1,82 @@
...
+                    !describeConfigurationSettings.data ||
+                    !describeConfigurationSettings.data.ConfigurationSettings) {
+                    helpers.addResult(results, 3,
+                        'Unable to query for environment configuration settings',
+                        region, resource);   
+                    return ecb();
+                }
+
+                if (!describeConfigurationSettings.data.ConfigurationSettings.length) {
+                    helpers.addResult(results, 0, 'No environment configuration settings found', region, resource);
```
matthewdfuller, line 51: Shouldn't this case be a fail? If there are no environment settings, then managed updates can't be enabled.
```suggestion
                    helpers.addResult(results, 2, 'No environment configuration settings found', region, resource);
```

plugins/aws/elasticbeanstalk/managedPlatformUpdates.js
```diff
@@ -0,0 +1,82 @@
...
+                    var param = describeConfigurationSettings.data.ConfigurationSettings[p];
+
+                    for (var s in param.OptionSettings) {
+                        var setting = param.OptionSettings[s];
+
+                        if (setting.Namespace && setting.Namespace === 'aws:elasticbeanstalk:managedactions' &&
+                            setting.OptionName && setting.OptionName === 'ManagedActionsEnabled') {
+                            if (setting.Value && setting.Value === 'true') {
+                                helpers.addResult(results, 0,
+                                    'Managed platform updates for environment: ' + environment.EnvironmentName + ' are enabled',
```
matthewdfuller, line 65: I think you should also keep a boolean here to track if a result was produced. Something like:
```
var found = false;
....
if (...) //matches {
    found = true;
}
```

And then at the end, after the for loop, you can do:
```
if (!found) {
    helpers.addResult(results, 2,
                                    'Managed platform updates for environment: ' + environment.EnvironmentName + ' are not enabled',
                                    region, resource);
}
```

plugins/aws/elasticbeanstalk/managedPlatformUpdates.js
AkhtarAmir: Yes you are right

giorod3 (approved):

collectors/aws/elasticbeanstalk/describeConfigurationSettings.js
```diff
@@ -0,0 +1,24 @@
...
+module.exports = function(AWSConfig, collection, callback) {
+    var elasticbeanstalk = new AWS.ElasticBeanstalk(AWSConfig);
+
+    async.eachLimit(collection.elasticbeanstalk.describeEnvironments[AWSConfig.region].data, 15, function(environment, cb) {
+        var params = {
+            ApplicationName: environment.ApplicationName,
+            EnvironmentName: environment.EnvironmentName
+        };
+        elasticbeanstalk.describeConfigurationSettings(params, function(err, data) {
+            collection.elasticbeanstalk.describeConfigurationSettings[AWSConfig.region][environment.EnvironmentName] = {};
```
matthewdfuller, line 13: @giorod3 @AkhtarAmir - I am concerned about using the environment name here as the unique key. Different projects can share the same environment name. For example, I can have the following:
```
Project: xyz
Environment: prod

Project: abc
Environment: prod
```

In this case, the second will overwrite the first. I think we need to either use a full ARN or some combination of app name + environment name,

AkhtarAmir: Yes ARN makes more sense. Fixed

PR #280, opened 2020-08-23
Feature/39 | Added Autoscaling Active Notification plugin, spec files

AkhtarAmir:


plugins/aws/autoscaling/asgActiveNotifications.js
```diff
@@ -0,0 +1,70 @@
...
+            }
+
+            async.each(describeAutoScalingGroups.data, function(asg, cb){
+                var resource = asg.AutoScalingGroupARN;
+                var notificationConfiguration = helpers.addSource(cache, source,
+                    ['autoscaling', 'describeNotificationConfigurations', region, asg.AutoScalingGroupARN]);
+
+                if (!notificationConfiguration) return cb();
+
+                if (notificationConfiguration.err || !notificationConfiguration.data) {
```
matthewdfuller, line 43: When doing a nested check for the data, these should be combined so that we always return a result for every ASG:
```suggestion
                if (!notificationConfiguration || notificationConfiguration.err || !notificationConfiguration.data) {
```

matthewdfuller (changes requested):
I checked the AWS docs and they use `Auto Scaling`, so we should be consistent. Please update all cases of `Autoscaling`, `AutoScaling`, `autoscaling`, etc. to `Auto Scaling`.

AkhtarAmir:
> I checked the AWS docs and they use `Auto Scaling`, so we should be consistent. Please update all cases of `Autoscaling`, `AutoScaling`, `autoscaling`, etc. to `Auto Scaling`.

I have updated all the occurrences  as suggested.

PR #286, opened 2020-08-29
Feature/57 | Added CloudTrail Object Lock plugin, spec files

AkhtarAmir:


plugins/aws/cloudtrail/cloudtrailObjectLock.js
```diff
@@ -0,0 +1,81 @@
...
+
+                if (!getObjectLockConfiguration) {
+                    helpers.addResult(results, 3,
+                        'Error querying for object lock configuration for bucket: ' + trail.S3BucketName + ': ' + helpers.addError(getObjectLockConfiguration),
+                        region, resource);
+
+                    return cb();
+                }
+
+                if (getObjectLockConfiguration.err || !getObjectLockConfiguration.data) {
```
giorod3, line 54: `getObjectLockConfiguration.err` should be added to the error check above (line 46)

AkhtarAmir: So if the object lock is not enabled for s3 bucket, getObjectLockConfiguration API sends an error in response, not an empty data object. Which means that object lock is not enabled for the bucket and we should add the failed result.

giorod3 (approved):

giorod3 (changes requested):
@AkhtarAmir Lets sync on Monday about this as well

giorod3 (approved):

PR #290, opened 2020-08-30
feature/15 | Added IAM Group Inline Policies plugin, spec file

AkhtarAmir:


plugins/aws/iam/groupInlinePolicies.js
```diff
@@ -0,0 +1,63 @@
...
+            }
+            
+            if (listGroupPolicies.data.PolicyNames && !listGroupPolicies.data.PolicyNames.length) {
+                helpers.addResult(results, 0,
+                    'Group: ' + group.GroupName + ' does not contain any inline policy',
+                    'global', group.Arn);
+            }else {
+                helpers.addResult(results, 2,
+                    'Group: ' + group.GroupName + ' contains ' + listGroupPolicies.data.PolicyNames.length + ' inline policy(s)',
+                    'global', group.Arn);
```
matthewdfuller, line 55: It's possible this could break if `listGroupPolicies.data.PolicyNames` is not set. The first `if` checks that `listGroupPolicies.data.PolicyNames` exists, so if it doesn't it will go to the `else`, but then inside the else it attempts to access `listGroupPolicies.data.PolicyNames.length` (which might not exist and throw an undefined error).

I think we should reorder these:
```
if (!listGroupPolicies.data.PolicyNames || !listGroupPolicies.data.PolicyNames.length) {
    // PASS condition
} else {
    // FAIL condition
}

giorod3 (approved):

matthewdfuller (approved):

PR #291, opened 2020-08-30
feature/11 | Added IAM Cross Account Access plugin,spec file

AkhtarAmir:


plugins/aws/iam/crossAccountAccess.js
```diff
@@ -0,0 +1,64 @@
...
+            if (!role.AssumeRolePolicyDocument) {
+                return;
+            }
+
+            var assumeRolePolicy = JSON.parse(decodeURIComponent(role.AssumeRolePolicyDocument));
+            
+            if (assumeRolePolicy.Statement &&
+                assumeRolePolicy.Statement.length &&
+                assumeRolePolicy.Statement[0].Principal &&
+                assumeRolePolicy.Statement[0].Principal.AWS) {
```
matthewdfuller, line 44: We need to check all of the statements, not just `[0]`, so this should be a loop for this section.

plugins/aws/iam/crossAccountAccess.js
```diff
@@ -0,0 +1,64 @@
+var helpers = require('../../../helpers/aws');
+
+module.exports = {
+    title: 'Cross Account Access',
+    category: 'IAM',
+    description: 'Ensures that either MFA or external IDs are used to access AWS resources.',
```
matthewdfuller, line 6:
```suggestion
    description: 'Ensures that either MFA or external IDs are used to access AWS roles.',
```

plugins/aws/iam/crossAccountAccess.js
```diff
@@ -0,0 +1,64 @@
+var helpers = require('../../../helpers/aws');
+
+module.exports = {
+    title: 'Cross Account Access',
```
matthewdfuller, line 4:
```suggestion
    title: 'Cross-Account Access External ID and MFA',
```

plugins/aws/iam/crossAccountAccess.js
```diff
@@ -0,0 +1,64 @@
...
+    category: 'IAM',
+    description: 'Ensures that either MFA or external IDs are used to access AWS resources.',
+    more_info: 'MFA/external ID adds an extra layer of security on top of roles, temporary security credentials and facilitates external third-party accounts to access your AWS resources in a secure way.',
```
matthewdfuller, line 7:
```suggestion
    more_info: 'IAM roles should be configured to require either a shared external ID or use an MFA device when assuming the role.',
```

plugins/aws/iam/crossAccountAccess.js
```diff
@@ -0,0 +1,64 @@
...
+    link: 'https://aws.amazon.com/blogs/aws/mfa-protection-for-cross-account-access/',
+    recommended_action: 'Edit IAM role to require MFA or external ID.',
```
matthewdfuller, line 9:
```suggestion
    recommended_action: 'Update the IAM role to either require MFA or use an external ID.',
```

exports.js
```diff
@@ -113,6 +113,7 @@ module.exports = {
         'accessKeysLastUsed'            : require(__dirname + '/plugins/aws/iam/accessKeysLastUsed.js'),
         'accessKeysRotated'             : require(__dirname + '/plugins/aws/iam/accessKeysRotated.js'),
         'certificateExpiry'             : require(__dirname + '/plugins/aws/iam/certificateExpiry.js'),
+        'crossAccountAccess'            : require(__dirname + '/plugins/aws/iam/crossAccountAccess.js'),
```
matthewdfuller, line 116:
```suggestion
        'crossAccountMfaExtIdAccess'            : require(__dirname + '/plugins/aws/iam/crossAccountMfaExtIdAccess.js'),
```

Also, please update the plugin file name

plugins/aws/iam/crossAccountAccess.js
AkhtarAmir: So I created a role and as per my understanding, these settings are role-based. I couldn't find a scenario where there will be more than once statements against one role policy document.
Capture

AkhtarAmir: And if there are multiple statements, should we check that at least one should have MFA/external ID enabled or all statements?

giorod3: Please check that each statement has either mfa or external ID enabled

giorod3 (approved):

giorod3 (changes requested):
@AkhtarAmir Lets sync on this one on Monday.

giorod3 (approved):

PR #298, opened 2020-09-01
Feature/plugin/credential report collection fix

jchrisfarris:
Increases number of retries on credential report collection to reduce number of false-positive UNKNOWN findings.


AkhtarAmir:
This can be ignored for now because it can cause the execution time issue.

PR #299, opened 2020-09-01
Feature/plugin/lambda log group fix

jchrisfarris:
Checks for lambda@edge log groups, their log group naming is inconsistent with regular Lambda


AkhtarAmir:
Added separate PR for this, can be closed

PR #300, opened 2020-09-01
Feature/plugin/ebs default ebs encryption

jchrisfarris:
New plugin to detect if the EBS Default Encryption setting is enabled in all regions. 


AkhtarAmir:
Added separate PR for this, Can be closed.

PR #301, opened 2020-09-03
feature/69 | Added Unused EBS Volumes plugin, spec file

AkhtarAmir:


plugins/aws/ec2/ebsUnusedVolumes.js
```diff
@@ -0,0 +1,78 @@
...
+            }
+
+            var unusedEbsVolumeFound = false;
+            describeVolumes.data.forEach(function(volume) {
+                if (volume.VolumeId) {
+                    if(!usedEbsVolumes.includes(volume.VolumeId)) {
+                        unusedEbsVolumeFound = true;
+                        helpers.addResult(results, 2,
+                            'EBS Volume is not attached to any EC2 instance',
+                            region, volume.VolumeId);
```
giorod3, line 64: can we replace `volume.VolumueId` with an ARN?

plugins/aws/ec2/ebsUnusedVolumes.js
```diff
@@ -0,0 +1,78 @@
...
+                    }
+                }
+            });
+
+            if(!unusedEbsVolumeFound){
```
giorod3, line 69: instead of one passing result, can we move the passing result for each volume to be the else statement of the failing result? 
`  if(!usedEbsVolumes.includes(volume.VolumeId)) {
                        unusedEbsVolumeFound = true;
                        helpers.addResult(results, 2,
                            'EBS Volume is not attached to any EC2 instance',
                            region, volume.VolumeId);
                     } else {
                     ** ADD PASSING RESULT HERE **
                     }
`

giorod3 (changes requested):

plugins/aws/ec2/ebsUnusedVolumes.js
AkhtarAmir: Apparently we can't. EBS Volume doesn't have an ARN.

giorod3: Can you recreate it using the available information and an arn template like this: `arn:aws:ec2:us-east-1:123456789012:volume/{VolumeId}`

AkhtarAmir: Sure. Will do that.

giorod3:
Theres a failing check, can you fix it so we can merge this pr

giorod3 (approved):

PR #302, opened 2020-09-03
feature/75 | Added EC2 Unassociated Elastic IP address plugin, spec file

AkhtarAmir:


plugins/aws/ec2/unassociatedElasticIp.js
```diff
@@ -0,0 +1,52 @@
...
+            describeAddresses.data.forEach(function(elasticIp){
+                if(!elasticIp.AssociationId) {
+                    unassociatedElasticIpFound = true;
+                    helpers.addResult(results, 2, 'Elastic IP address is not associated to any resource', region, elasticIp.AllocationId);
+                }
+            });
+            
+            if(!unassociatedElasticIpFound){
+                helpers.addResult(results, 0, 'All Elastic IP addresses are associated to resources', region);
+            }
```
giorod3, line 45: @matthewdfuller for EIP should we be returning a pass/fail for each EIP or will this be ok?

matthewdfuller: From call: each one should have its own result

giorod3 (changes requested):
@AkhtarAmir This pull request is failing lint

giorod3 (approved):

PR #305, opened 2020-09-05
feature/10 | Added IAM Trusted Cross-Account Roles plugin, spec file

AkhtarAmir:


plugins/aws/iam/trustedCrossAccountRoles.js
```diff
@@ -0,0 +1,93 @@
...
+    more_info: 'IAM roles should be configured to allow access to trusted account IDs.',
+    link: 'https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_common-scenarios_aws-accounts.html',
+    recommended_action: 'Delete the IAM roles that are associated with untrusted account IDs.',
+    apis: ['IAM:listRoles'],
+    settings: {
+        whitelisted_accounts: {
+            name: 'Allowed cross-account role account IDs',
+            description: 'Return a failing result if cross-account role contains any account ID other than these account IDs',
+            regex: '^[0-9{12}]$',
+            default: [ '123456654321' ]
```
giorod3, line 16: make this a comma separated string

plugins/aws/iam/trustedCrossAccountRoles.js
```diff
@@ -0,0 +1,93 @@
...
+            helpers.addResult(results, 0, 'No IAM roles found');
+            return callback(null, results, source);
+        }
+
+        var awsRolesFound = false;
+
+        listRoles.data.forEach(role => {
+            if (!role.AssumeRolePolicyDocument) {
+                return;
+            }
```
giorod3, line 53: you can simplify this 
```suggestion
            if (!role.AssumeRolePolicyDocument) return;
```

plugins/aws/iam/trustedCrossAccountRoles.js
```diff
@@ -0,0 +1,93 @@
...
+
+            var assumeRolePolicy = JSON.parse(decodeURIComponent(role.AssumeRolePolicyDocument));
```
giorod3, line 55: wrap this into a try/catch in case the parse fails

giorod3 (changes requested):

giorod3 (approved):

plugins/aws/iam/trustedCrossAccountRoles.js
```diff
@@ -0,0 +1,96 @@
...
+    title: 'Trusted Cross Account Roles',
+    category: 'IAM',
+    description: 'Ensures that only trusted cross-account IAM roles can be used.',
+    more_info: 'IAM roles should be configured to allow access to trusted account IDs.',
+    link: 'https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_common-scenarios_aws-accounts.html',
+    recommended_action: 'Delete the IAM roles that are associated with untrusted account IDs.',
+    apis: ['IAM:listRoles'],
+    settings: {
+        whitelisted_accounts: {
+            name: 'Allowed cross-account role account IDs',
```
matthewdfuller, line 13:
```suggestion
            name: 'Allowed Cross-Account Role Account IDs',
```

plugins/aws/iam/trustedCrossAccountRoles.js
```diff
@@ -0,0 +1,96 @@
...
+            description: 'Return a failing result if cross-account role contains any account ID other than these account IDs',
+            regex: '^[0-9{12}]$',
```
matthewdfuller, line 15: This should be extended so that the commas can be handled. For example:
Valid: `123456789101`
Valid: `123456789101,123456789102,123456789103`
NOT Valid: `123456789101,`
NOT Valid: `123456789101,,,`
```suggestion
            regex: '^[0-9{12}]$',
```

plugins/aws/iam/trustedCrossAccountRoles.js
```diff
@@ -0,0 +1,96 @@
...
+            if (!role.AssumeRolePolicyDocument) return;
+
+            try {
+                var assumeRolePolicy = JSON.parse(decodeURIComponent(role.AssumeRolePolicyDocument));
+            } catch (e) {
+                helpers.addResult(results, 3,
+                    'IAM role policy document is not valid JSON.',
+                    'global', role.Arn);
+                return;
+            }
```
matthewdfuller, line 58: This can be replaced using the AWS helper `normalizePolicyDocument` which will handle the try/catch + the normalization of the effect/statement/resource to `[]`

plugins/aws/iam/trustedCrossAccountRoles.js
```diff
@@ -0,0 +1,96 @@
...
+                if (crossAccountRole && !restrictedAccounts.length) {
+                    helpers.addResult(results, 0,
+                        'Cross-account role :' + role.RoleName + ': contains trusted account ID(s)',
+                        'global', role.Arn);
+                }
+                else if (crossAccountRole) {
+                    helpers.addResult(results, 2,
+                        'Cross-account role :' + role.RoleName + ': contains these untrusted account ID(s): ' + restrictedAccounts.join(', '),
+                        'global', role.Arn);
+                }
```
matthewdfuller, line 86: Passing condition is needed here for `IAM role does not contain cross-account statements`

plugins/aws/iam/trustedCrossAccountRoles.js
```diff
@@ -0,0 +1,96 @@
...
+            }
```
matthewdfuller, line 87: Passing condition is needed here for `IAM role does not contain trust relationship statements`.

matthewdfuller (changes requested):

plugins/aws/iam/trustedCrossAccountRoles.js
AkhtarAmir: So I am displaying results only for cross-account roles not for all IAM roles so I don't think this check is required.

plugins/aws/iam/trustedCrossAccountRoles.js
AkhtarAmir: Again, the result will only be displayed if the role is cross-account (having cross-account principal) and it has trust relationship statements. If we don't find any such cross-account, a result is being displayed at the end that no cross-account role found.

helpers/aws/functions.js
```diff
@@ -194,10 +194,11 @@ function globalPrincipal(principal) {
     return false;
 }
 
-function crossAccountPrincipal(principal, accountId) {
+function crossAccountPrincipal(principal, accountId, fetchPrincipals = false) {
```
giorod3, line 197:
```suggestion
function crossAccountPrincipal(principal, accountId, fetchPrincipals) {
```

helpers/aws/functions.js
```diff
@@ -542,5 +548,5 @@ module.exports = {
     remediatePlugin: remediatePlugin,
     nullArray: nullArray,
     remediatePasswordPolicy:remediatePasswordPolicy,
-    remediateOpenPorts: remediateOpenPorts
+    remediateOpenPorts: remediateOpenPorts,
```
giorod3, line 551:
```suggestion
    remediateOpenPorts: remediateOpenPorts
```

giorod3 (changes requested):
change the result to pass per role not just per cross account role

plugins/aws/iam/trustedCrossAccountRoles.js
```diff
@@ -0,0 +1,82 @@
...
+            if (crossAccountRole && !restrictedAccountPrincipals.length) {
+                helpers.addResult(results, 0,
+                    `Cross-account role "${role.RoleName}" contains trusted account pricipals`,
+                    'global', role.Arn);
+            }
+            else if (crossAccountRole) {
+                helpers.addResult(results, 2,
+                    `Cross-account role "${role.RoleName}" contains these untrusted account principals: ${restrictedAccountPrincipals.join(', ')}`,
+                    'global', role.Arn);
+            }
```
dipsubha06, line 73: Do we not need any else condition?

AkhtarAmir:
This PR has been split and regenerated

PR #306, opened 2020-09-05
feature/94 | Added ELBv2 Deletion Protection plugin, spec file

AkhtarAmir:


plugins/aws/elbv2/elbv2DeletionProtection.js
```diff
@@ -0,0 +1,86 @@
...
+                    helpers.addResult(results, 3,
+                        'Unable to query for Application/Network load balancer attributes: ' +  helpers.addError(elbv2Attributes),
+                        region, resource);
+                    return cb();
+                }
+
+                if(!elbv2Attributes.data.Attributes || !elbv2Attributes.data.Attributes.length){
+                    helpers.addResult(results, 2,
+                        'Application/Network load balancer attributes not found ' +  helpers.addError(elbv2Attributes),
+                        region, resource);
```
giorod3, line 53: this result is for unknown results, change the result to be 3 or change the message to exclude the helpers.addError

plugins/aws/elbv2/elbv2DeletionProtection.js
```diff
@@ -0,0 +1,86 @@
...
+                    return cb();
+                }
+
+                var found = false;
+
+                elbv2Attributes.data.Attributes.forEach(attribute => {
+                    if (attribute.Key && attribute.Key === 'deletion_protection.enabled') {
+                        found = true;
+                        if (attribute.Value === 'true') {
```
giorod3, line 62: check attribute.Value before doing the comparison 
```suggestion
                        if (attribute.Value && attribute.Value === 'true') {
```

plugins/aws/elbv2/elbv2DeletionProtection.js
```diff
@@ -0,0 +1,86 @@
...
+                        } else {
+                            helpers.addResult(results, 2,
+                                'Load balancer :' + elb.LoadBalancerName + ': does not have deletion protection enabled',
+                                region, resource);
+                        }
+                    }
+                });
+
+                if (!found) {
+                    helpers.addResult(results, 2, 'Deletion protection key not found', region, resource);
```
giorod3, line 75:
```suggestion
                    helpers.addResult(results, 2, 'Deletion protection not found', region, resource);
```

giorod3 (changes requested):

giorod3 (approved):

PR #311, opened 2020-09-08
feature/100 | Added EMR Encryption In Transit plugin, spec file

AkhtarAmir:
Fixed description, change from FTP to SSH service.


AkhtarAmir:
Will add encryption at rest plugin as well and create PR again

Issue #269, opened 2020-08-12
Configuration value "project" not set properly when using the configuration method 'OPTION 2' for Google


lhankinsgdl:
When passing in Google configuration information via 'OPTION 1', we manually set `"project": "your-project-name",`.

When using 'OPTION 3', `project` is set properly using either the environment variable:
   `project: process.env.GOOGLE_PROJECT_ID,`
... or via the file pulled from `process.env.GOOGLE_APPLICATION_CREDENTIALS`:
   `GoogleConfig.project = GoogleConfig.project_id;`

But when using 'OPTION 2' with a .json file downloaded from the Google control panel after creating the service account, `project_id` exists in the json but not `project`, so `GoogleConfig.project` never gets set.

This causes all the checks to appear as "UNKNOWN".

A workaround is to add the `project` item in the .json file, setting it to match the `project_id` node.

lhankinsgdl:
This has changed with v2.0.  Similar issue, but now the fix is to add:
`cloudConfig.project = cloudConfig.project_id;`
...after the config file is parsed.

PR #315, opened 2020-09-11
Fixing gcp json parsing for #269

lhankinsgdl:
This should solve #269 by copying the `project_id` variable (which Google writes to the .json file) to `project`, which cloudsploit expects.


giorod3 (approved):

giorod3:
@lhankinsgdl thank you for this contribution

PR #334, opened 2020-09-21
Add master/manager role check

coj337:
There's a common CIS check needed by orgs, see: https://www.cloudconformity.com/knowledge-base/aws/IAM/master-and-manager-role.html

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ vscode/*
 coverage/
 .nyc_output
 config.js
+.env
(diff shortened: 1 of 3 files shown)


giorod3:
@coj337 Thank you for this contribution. we have merged this internally and renamed it iamMasterManagerRoles.js
https://github.com/aquasecurity/cloudsploit/blob/master/plugins/aws/iam/iamMasterManagerRoles.js

PR #339, opened 2020-09-22
Sploit 716 | Added plugin to compare node pool k8 version with that of cluster and the spec file

dipsubha06:
Added the following library -
 nyc - for test coverage 
 compare-versions - for comparing versions.
Below is the test coverage report
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,6 @@
     "aws-sdk": "^2.740.0",
     "azure-storage": "^2.10.3",
     "csv-write-stream": "^2.0.0",
-    "compare-versions": "^3.6.0",
     "fast-safe-stringify": "^2.0.6",
     "googleapis": "^40.0.1",
     "minimatch": "^3.0.4",
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "AWS, Azure, GCP, Oracle, GitHub security scanning scripts",
   "main": "index.js",
   "scripts": {
-    "test": "nyc --reporter=text mocha './**/*.spec.js'",
+    "test": "mocha './**/*.spec.js'",
     "test-watch": "nodemon --exec npm run test",
     "test-cov": "nyc --reporter text --reporter cobertura --reporter lcov --all -- npm run test",
     "test-cov-watch": "nodemon --exec npm run test-cov",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -4,7 +4,7 @@
   "description": "AWS, Azure, GCP, Oracle, GitHub security scanning scripts",
   "main": "index.js",
   "scripts": {
-    "test": "mocha './**/*.spec.js'",
+    "test": "nyc --reporter=text mocha './**/*.spec.js'",
```
matthewdfuller, line 7: This is similar to the `test-cov` command below that already exists. Could we keep `test` as just for the mocha runs and leave `test-cov` for the coverage report?

plugins/azure/kubernetesservice/aksAgentVersion.js
```diff
@@ -0,0 +1,63 @@
...
+                var kubernetesVersion = managedCluster.kubernetesVersion;
+                var agentPoolProfiles = managedCluster.agentPoolProfiles;
+
+                if (!agentPoolProfiles || !agentPoolProfiles.length) {
+                    helpers.addResult(results, 3,
+                        'Unable to query for Kubernetes cluster node profile' , location, managedCluster.id);
+                } else {
+                    agentPoolProfiles.forEach(agentPoolProfile =>{
+                        if (agentPoolProfile.orchestratorVersion &&
+                            compareVersions(agentPoolProfile.orchestratorVersion,kubernetesVersion) === -1 ) {
```
matthewdfuller, line 46: Can you use the same comparison logic from the `aksVersion` plugin to avoid adding another package dependency?
(We try to avoid new packages unless they are absolutely required).

package.json
dipsubha06: I have removed the nyc related changes

plugins/azure/kubernetesservice/aksAgentVersion.js
dipsubha06: in aksversion we are just checking if there is any entry in updateprofile array to decide whether we are at latest version. We are not checking any version.But here we are checking version of the node pool with the aks cluster it self.

matthewdfuller: Got it - is it possible to extract the relevant functionality from the package and put it in `helpers/azure.js` instead? Want to avoid adding any dependencies if we can.

dipsubha06: I have wrote a simple function for this and added in helpers/shared.js as this is not specific to azure. Please take a look

PR #342, opened 2020-09-26
Hotfix + Spec | IAM SSH Keys Rotated plugin

AkhtarAmir:


plugins/aws/iam/sshKeysRotated.js
```diff
@@ -93,7 +93,7 @@ module.exports = {
         }
 
         if (!found) {
-            helpers.addResult(results, 0, 'No SSH keys found');
+            helpers.addResult(results, 0, 'No SSH keys found', 'global', obj.arn);
```
giorod3, line 96: you can't reference obj its out of scope.

giorod3 (changes requested):

plugins/aws/iam/sshKeysRotated.js
AkhtarAmir: Oh yes you are right. In that case we shouldn't pass any resource to addResult.

giorod3 (approved):

matthewdfuller (approved):

PR #345, opened 2020-09-26
Feature/29 | Added AutoScaling ELB Health Check Active plugin, spec file

AkhtarAmir:


giorod3 (approved):

matthewdfuller (changes requested):
Overall, the code is good, but we need to check one extra use case when `LoadBalancerNames` in the response is empty (which indicates the ASG does not have an ELB associated with it).

If `LoadBalancerNames` is not set (or an empty array), then we should mark the ASG as `PASS` with the result `ASG does not use ELBs`.

giorod3:
@AkhtarAmir this PR is failing lint, it looks like an indent issue. Can you fix this so I can approve it

plugins/aws/autoscaling/elbHealthCheckActive.js
```diff
@@ -0,0 +1,63 @@
...
+                        helpers.addResult(results, 0,
+                            `Auto Scaling group "${asg.AutoScalingGroupName}" does not use ELBs`,
+                            region, resource);
+                            return;
+                    } else {
+                        helpers.addResult(results, 0,
+                            `Auto Scaling group "${asg.AutoScalingGroupName}" has ELB health check active`,
+                            region, resource);
+                    }
+                } else {
```
giorod3, line 51: make an else if to ensure that the user is using elb here instead of inside of the other if statement

giorod3 (approved):

matthewdfuller (approved):

PR #351, opened 2020-09-27
feature/1152 | Added SQS Public Access plugin, spec file

AkhtarAmir:


plugins/aws/sqs/sqsPublicAccess.js
```diff
@@ -0,0 +1,107 @@
+var async = require('async');
+var helpers = require('../../../helpers/aws');
+
+module.exports = {
+    title: 'SQS Public Access',
+    category: 'SQS',
+    description: 'Ensures that SQS queues are not publically accessible',
```
giorod3, line 7:
```suggestion
    description: 'Ensures that SQS queues are not publicly accessible',
```

plugins/aws/sqs/sqsPublicAccess.js
```diff
@@ -0,0 +1,107 @@
...
+    more_info: 'SQS queues should be not be publically accessible to prevent unauthorized actions.',
```
giorod3, line 8:
```suggestion
    more_info: 'SQS queues should be not be publicly accessible to prevent unauthorized actions.',
```

plugins/aws/sqs/sqsPublicAccess.js
```diff
@@ -0,0 +1,107 @@
...
+
+            if (!listQueues) return rcb();
+
+            if (listQueues.err) {
+                helpers.addResult(results, 3,
+                    `Unable to query for SQS queues: ${helpers.addError(listQueues)}`, region);
+                return rcb();
+            }
+
+            if (!listQueues.data || !listQueues.data.length) {
```
giorod3, line 35: move listQueues.data to the error check above

plugins/aws/sqs/sqsPublicAccess.js
```diff
@@ -0,0 +1,107 @@
...
+                    var policy = JSON.parse(getQueueAttributes.data.Attributes.Policy);
+                } catch (e) {
+                    helpers.addResult(results, 3,
+                        'The SQS queue policy is not valid JSON.',
+                        region, resource);
+
+                    return cb();
+                }
+
+                var statements = helpers.normalizePolicyDocument(policy);
```
giorod3, line 75: lets sync on this, there are a few places where we can just use the helpers.normalizePolicyDocument instead of having the try catch

plugins/aws/sqs/sqsPublicAccess.js
```diff
@@ -0,0 +1,107 @@
...
+                        statement.Effect === 'Allow' &&
+                        statement.Principal &&
+                        statement.Principal === '*') {
+                        publicStatements.push(statement);
+                    }
+                }
+
+                if (!publicStatements.length) {
+                    helpers.addResult(results, 0,
+                        `SQS queue "${queueName}" is not publically accessible`,
```
giorod3, line 91:
```suggestion
                        `SQS queue "${queueName}" is not publicly accessible`,
```

plugins/aws/sqs/sqsPublicAccess.js
```diff
@@ -0,0 +1,107 @@
...
+                        region, resource);
+                } else {
+                    helpers.addResult(results, 2,
+                        `SQS queue "${queueName}" is publically accessible`,
```
giorod3, line 95:
```suggestion
                        `SQS queue "${queueName}" is publicly accessible`,
```

giorod3 (changes requested):

plugins/aws/sqs/sqsPublicAccess.js
AkhtarAmir: In this particular case, if there are no SQS queues, the API returns nothing and listQueues is just an empty array. It is implemented same in other sqs plugins as well

plugins/aws/sqs/sqsPublicAccess.js
AkhtarAmir: Yeah I tried this using only helpers.normalizePolicyDocument and it works. So I am changing it for you to review and if you still think that it needs discussion, we will discuss it in meeting

giorod3 (approved):

giorod3 (approved):

plugins/aws/sqs/sqsPublicAccess.js
```diff
@@ -0,0 +1,97 @@
...
+                var statements = helpers.normalizePolicyDocument(getQueueAttributes.data.Attributes.Policy);
+
+                var publicStatements = [];
+                for (var s in statements) {
+                    var statement = statements[s];
+
+                    if (statement.Effect &&
+                        statement.Effect === 'Allow' &&
+                        statement.Principal &&
+                        statement.Principal === '*') {
```
matthewdfuller, line 74: Can use the helper globalPrincipal function here

matthewdfuller (changes requested):

giorod3 (approved):

PR #361, opened 2020-10-08
Hotfix: Azure keyvaultsRecoveryEnabled, keyExpirationEnabled

AkhtarAmir:


giorod3 (approved):

plugins/azure/keyvaults/keyExpirationEnabled.js
```diff
@@ -51,7 +51,7 @@ module.exports = {
                         
                         if (key.attributes) {
                             let attributes = key.attributes;
-                            if (attributes.expires && attributes.expires !== null && attributes.expires !== '') {
+                            if (attributes.exp && attributes.exp !== null && attributes.exp !== '') {
```
matthewdfuller, line 54:
```suggestion
                            if ((attributes.expires && attributes.expires !== null && attributes.expires !== '') || (attributes.exp && attributes.exp !== null && attributes.exp !== '')) {
```

matthewdfuller: Just to keep it backwards compatible in case all the Azure APIs haven't upgraded

PR #364, opened 2020-10-10
Chore: Bumped to node 14

ducksecops:
Node 8 was EOL as off 31st December 2019. Node 14 is the current LTS branch and is supported until 30th April 2023


matthewdfuller:
Thanks for the change @ducksecops - this should be resolved by https://github.com/aquasecurity/cloudsploit/pull/366

PR #358, opened 2020-10-03
Bugfix/plugins/google/logging/: fix status of projectOwnershipLogging when log metrics not found

lborguetti:
Hi guys,

Currently when I run the `projectOwnershipLogging` plugin it returns ok when I do not have log metrics or alert policies. 

I think the correct thing would be to return fail, according to the description of the plugin: [here](https://github.com/aquasecurity/cloudsploit/blob/master/plugins/google/logging/projectOwnershipLogging.js#L7).

```
$ ./index.js --config=./config.js --plugin projectOwnershipLogging

   _____ _                 _  _____       _       _ _
  / ____| |               | |/ ____|     | |     (_) |
 | |    | | ___  _   _  __| | (___  _ __ | | ___  _| |_
 | |    | |/ _ \| | | |/ _` |\___ \| '_ \| |/ _ \| | __|
 | |____| | (_) | |_| | (_| |____) | |_) | | (_) | | |_
  \_____|_|\___/ \__,_|\__,_|_____/| .__/|_|\___/|_|\__|
                                   | |
                                   |_|

  CloudSploit by Aqua Security, Ltd.
  Cloud security auditing for AWS, Azure, GCP, Oracle, and GitHub

INFO: Using CloudSploit config file: ./config.js
INFO: Skipping AWS pagination mode
INFO: Testing plugin: Project Ownership Logging
INFO: Determining API calls to make...
INFO: Found 2 API calls to make for google plugins
INFO: Collecting metadata. This may take several minutes...
INFO: Metadata collection complete. Analyzing...
INFO: Analysis complete. Scan report to follow...

  ┌────────┬────────────────┬────────────────────────────────────────────────────────┬──────┬────┬────────┬─────────────┐
  │ Catego │     Plugin     │                      Description                       │ Reso │ Re │ Status │   Message   │
  │   ry   │                │                                                        │ urce │ gi │        │             │
  │        │                │                                                        │      │ on │        │             │
  ├────────┼────────────────┼────────────────────────────────────────────────────────┼──────┼────┼────────┼─────────────┤
  │ Loggin │ Project        │ Ensures that logging and log alerts exist for project  │ N/A  │ gl │ OK     │ No log      │
  │ g      │ Ownership      │ ownership assignments and changes                      │      │ ob │        │ metrics     │
  │        │ Logging        │                                                        │      │ al │        │ found       │
  └────────┴────────────────┴────────────────────────────────────────────────────────┴──────┴────┴────────┴─────────────┘
INFO: Scan complete
```

[...]

PR #386, opened 2020-10-20
Bugfix/plugins/google/logging/: fix status when log metrics or alert policies not found

lborguetti:
Hi guys,

Currently when I run the logging plugins it returns ok when I do not have log metrics or alert policies.

Same as https://github.com/aquasecurity/cloudsploit/pull/358

This pull request changes the Status to FAIL when log metrics or alert policies not found in the project.

If it is better I can open a PR for each plugin.

Regards,

### DIFF
--- a/plugins/google/logging/auditConfigurationLogging.js
+++ b/plugins/google/logging/auditConfigurationLogging.js
@@ -43,12 +43,12 @@ module.exports = {
             }
 
             if (!metrics.data.length > 0) {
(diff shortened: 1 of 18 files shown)


matthewdfuller:
@lborguetti - thanks for this; good change!

lborguetti:
> @lborguetti - thanks for this; good change!

Thanks @matthewdfuller

PR #396, opened 2020-10-26
Feature/ak-69: Added AWS EC2 Unused Virtual Private Gateway plugin, test cases

AkhtarAmir:


plugins/aws/ec2/unusedVirtualPrivateGateway.js
```diff
@@ -0,0 +1,68 @@
...
+                return rcb();
+            }
+
+            describeVpnGateways.data.forEach(function(vpn){
+                var resource = `arn:${awsOrGov}:vpc:${region}:${accountId}:vpn-gateway/${vpn.VpnGatewayId}`;
+                var vpnAttached = false;
+                if (vpn.VpcAttachments && vpn.VpcAttachments) {
+                    vpn.VpcAttachments.forEach(function(attachment){
+                        if(attachment.State && attachment.State === 'attached') {
+                            vpnAttached = true;
```
dipsubha06, line 47: Minor: cant we break here after we find there is some attachment  in the VPG? all we want to see is if the vpg is in use and if it has at least one attachment that means it is in use. hence no need to traverse the entire list.

dipsubha06 (approved):

plugins/aws/ec2/unusedVirtualPrivateGateway.js
```diff
@@ -0,0 +1,68 @@
...
+            if (!describeVpnGateways.data.length) {
+                helpers.addResult(results, 0,
+                    'No Virtual Private Gateways found', region);
+                return rcb();
+            }
+
+            describeVpnGateways.data.forEach(function(vpn){
+                var resource = `arn:${awsOrGov}:vpc:${region}:${accountId}:vpn-gateway/${vpn.VpnGatewayId}`;
+                var vpnAttached = false;
+                if (vpn.VpcAttachments && vpn.VpcAttachments) {
```
giorod3, line 44:
```suggestion
                if (vpn.VpcAttachments && vpn.VpcAttachments.length) {
```

giorod3 (approved):

PR #420, opened 2020-11-18
hotfix + spec: Added missing test cases for AWS CloudTrail plugins | Fixed syntax errors

AkhtarAmir:


plugins/aws/cloudtrail/cloudtrailObjectLock.js
```diff
@@ -29,7 +29,7 @@ module.exports = {
             }
 
             if (!describeTrails.data.length) {
-                helpers.addResult(results, 0, 'No trails found', region);
+                helpers.addResult(results, 2, 'CloudTrails is not enabled', region);
```
giorod3, line 32:
```suggestion
                helpers.addResult(results, 0, 'No trails found', region);
```

giorod3: I am going to revert this

giorod3: actually, it will mess up the tests, can you revert this to No trails found, it sounds better.

giorod3 (changes requested):
just the one change and this is ready to go

plugins/aws/cloudtrail/cloudtrailObjectLock.js
AkhtarAmir: I had changed the test as well. The thing is that in all other CloudTrail plugins, we are displaying CloudTrail is not enabled. So I changed it to make it consistent. Do you still want me to change it back?

giorod3 (approved):

dipsubha06 (approved):

PR #424, opened 2020-11-23
SAAS-741: refactoring the remediation code

dipsubha06:
https://scalock.atlassian.net/browse/SAAS-741

fix for -
1) calling remediation in async way for multiple resources.
2) calling remediation for resources which are replicated over multiple region like cloudtrail.

### DIFF
--- a/engine.js
+++ b/engine.js
@@ -136,6 +136,7 @@ var engine = function(cloudConfig, settings) {
 
             plugin.run(collection, settings, function(err, results) {
                 if (!results.length) return console.log('ERROR: Nothing to report...');
+                var failedResources = [];
                 for (var r in results) {
                     // If we have suppressed this result, then don't process it
                     // so that it doesn't affect the return code.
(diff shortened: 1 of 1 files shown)


engine.js
```diff
@@ -160,22 +161,26 @@ var engine = function(cloudConfig, settings) {
                     // Add this to our tracking for the worst status to calculate
                     // the exit code
                     maximumStatus = Math.max(maximumStatus, results[r].status);
-                    // Remediation
+                    if (results[r].status === 2 && failedResources.indexOf(results[r].resource) === -1){
+                        failedResources.push(results[r].resource);
+                    }
+                }
+                async.forEach( failedResources,(resource, rcb) => {
                     if (settings.remediate && settings.remediate.length) {
```
giorod3, line 169: you should check `if (settings.remediate && settings.remediate.length) {` before going through the async loop, that way if the remediation is not set it won't even loop through those resources

giorod3 (changes requested):
just one comment

giorod3 (approved):

PR #433, opened 2020-11-29
feature/AKD-07: Added AWS Redshift Cluseter in VPC plugin and test cases

AkhtarAmir:


dipsubha06 (approved):

helpers/aws/functions.js
```diff
@@ -3,6 +3,10 @@ var regions = require(__dirname + '/regions');
 var AWS = require('aws-sdk');
 var helpers = require('../shared.js');
 
+function isValid(value) {
+    return (value !== undefined && value !== '' && value !== null);
+}
```
giorod3, line 8: this is unnecessary. you can just check (value && value.length) which will cover all these

giorod3 (changes requested):
just the one change

helpers/aws/functions.js
AkhtarAmir: You are right. I have removed this function

giorod3 (approved):

PR #441, opened 2020-12-04
feature-request: Added AWS CloudWatch Logs Log Retention Period plugin, test cases

AkhtarAmir:


plugins/aws/cloudwatchlogs/logRetentionPeriod.js
```diff
@@ -0,0 +1,66 @@
...
+    more_info: 'Retention settings can be used to specify how long log events are kept in CloudWatch Logs. Expired log events get deleted automatically.',
+    recommended_action: 'Ensure CloudWatch logs are retained for at least 90 days.',
+    link: 'https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/Working-with-log-groups-and-streams.html',
+    apis: ['CloudWatchLogs:describeLogGroups'],
+    settings: {
+        minimum_log_retention_period: {
+            name: 'CloudWatch Log Minimum Retention Period',
+            description: 'If set, CloudWatch Logs log groups should have a retention setting greater or equal to this value',
+            regex: '^[0-9]*$',
+            default: 0
```
giorod3, line 17: shouldn't the default be 90

plugins/aws/cloudwatchlogs/logRetentionPeriod.js
```diff
@@ -0,0 +1,66 @@
...
+        }
+    },
+
+    run: function(cache, settings, callback) {
+        var config = {
+            minimum_log_retention_period: settings.minimum_log_retention_period || this.settings.minimum_log_retention_period.default
+        };
+
+        if (config.minimum_log_retention_period === false) return callback();
```
giorod3, line 26: if the default is 0 wouldn't this callback? the plugin will never run. I don't think we should make this an opt-in plugin that has to be turned on. I think its a good plugin and should be on by default

giorod3 (changes requested):

plugins/aws/cloudwatchlogs/logRetentionPeriod.js
AkhtarAmir: Yes, that was the idea. But you are right. I've changed the default value to be 90

plugins/aws/cloudwatchlogs/logRetentionPeriod.js
```diff
@@ -0,0 +1,64 @@
...
+    more_info: 'Retention settings can be used to specify how long log events are kept in CloudWatch Logs. Expired log events get deleted automatically.',
+    recommended_action: 'Ensure CloudWatch logs are retained for at least 90 days.',
+    link: 'https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/Working-with-log-groups-and-streams.html',
+    apis: ['CloudWatchLogs:describeLogGroups'],
+    settings: {
+        minimum_log_retention_period: {
+            name: 'CloudWatch Log Minimum Retention Period',
+            description: 'If set, CloudWatch Logs log groups should have a retention setting greater or equal to this value',
+            regex: '^[0-9]*$',
+            default: 90
```
giorod3, line 17:
```suggestion
            default: '90'
```

giorod3 (changes requested):
just change to string and add logic

giorod3 (approved):

PR #442, opened 2020-12-07
feature/AK-67: Added AWS 'VPC Endpoint Cross Account Access' plugin and t…

AkhtarAmir:
…est cases


helpers/aws/functions.js
```diff
@@ -217,14 +218,19 @@ function crossAccountPrincipal(principal, accountId) {
         awsPrincipals = [awsPrincipals];
     }
 
+    var principals = [];
+    var response = false;
+
     for (var a in awsPrincipals) {
-        if (/^arn:aws:iam::[0-9]{12}.*/.test(awsPrincipals[a]) &&
+        if (/^arn:aws:iam|sts::[0-9]{12}.*/.test(awsPrincipals[a]) &&
```
giorod3, line 225: double check this regex. I tried it and it wasn't working this way. I added parenthesis and it worked

```suggestion
        if (/^arn:aws:(iam|sts)::[0-9]{12}.*/.test(awsPrincipals[a]) &&
```

AkhtarAmir: Yes, you are right. It was working that way as well but this way is better

helpers/aws/functions.js
```diff
@@ -217,14 +218,19 @@ function crossAccountPrincipal(principal, accountId) {
...
+        if (/^arn:aws:(iam|sts)::[0-9]{12}.*/.test(awsPrincipals[a]) &&
             awsPrincipals[a].indexOf(accountId) === -1) {
-            return true;
+            if (!fetchPrincipals) return true;
+            principals.push(awsPrincipals[a]);
         }
     }
 
-    return false;
+    if (fetchPrincipals) response = principals;
```
giorod3, line 232:
```suggestion
    if (fetchPrincipals) return principals;
```

helpers/aws/functions.js
```diff
@@ -217,14 +218,19 @@ function crossAccountPrincipal(principal, accountId) {
...
+    return response;
```
giorod3, line 233:
```suggestion
    return false;
```

giorod3 (changes requested):
just the minor changes

giorod3 (approved):

PR #443, opened 2020-12-07
feature-request: Added 'VPC Subnet Instances Present' plugin and test cases

AkhtarAmir:


plugins/aws/ec2/vpcSubnetInstancesPresent.js
```diff
@@ -0,0 +1,73 @@
...
+                });
+            }
+
+            describeSubnets.data.forEach(subnet => {
+                if(subnet.SubnetId && instanceSubnets[subnet.SubnetId]) {
+                    helpers.addResult(results, 0,
+                        `Subnet "${subnet.SubnetId}" has ${instanceSubnets[subnet.SubnetId]} instances attached`,
+                        region, subnet.SubnetArn);
+                } else {
+                    helpers.addResult(results, 2,
```
dipsubha06, line 62: this else can happen if somehow subnet.SubnetId is undefined ( which may be a rare case). In that case this addResult call also will fail as we are adding the subnet.SubnetId.

dipsubha06 (changes requested):

plugins/aws/ec2/vpcSubnetInstancesPresent.js
giorod3: @AkhtarAmir I agree with Subhadip.

plugins/aws/ec2/vpcSubnetInstancesPresent.js
```diff
@@ -0,0 +1,73 @@
...
+                        `Subnet "${subnet.SubnetId}" does not have any instance attached`,
+                        region, subnet.SubnetArn);
```
giorod3, line 64:
```suggestion
                    helpers.addResult(results, 2,
                        `Subnet does not have any instance attached`,
                        region, subnet.SubnetArn);
```

AkhtarAmir: Fixed

giorod3 (approved):

PR #466, opened 2020-12-17
feature/AKD-22: Added 'API Gateway Detailed CloudWatch Metrics' plugin and test cases

AkhtarAmir:


dipsubha06 (approved):

plugins/aws/apigateway/detailedCloudWatchMetrics.js
```diff
@@ -0,0 +1,80 @@
...
+                        'No Rest API Stages found',
+                        region, apiArn);
+                    return cb();
+                }
+
+                getStages.data.item.forEach(stage => {
+                    if (!stage.stageName) return;
+                    var stageArn = `arn:${awsOrGov}:apigateway:${region}::/restapis/${api.id}/stages/${stage.stageName}`;
+
+                    if (stage.methodSettings && stage.methodSettings['*/*'] && stage.methodSettings['*/*'].metricsEnabled) {
```
giorod3, line 61: is it always like this '*/*'?

AkhtarAmir: So this donates (resource_path)/(http_method). */* means each resource_path and each method. You can set the logs for some resources or for some methods in a resource. But I think for the plugin perspective, it should be */*

PR #469, opened 2020-12-21
Fix: Google DNSSecSigningAlgorithm Plugin Returns No Results When DNSSEC is Disabled

dave-pollock:
When DNSSEC is not enabled the managedZone object has no dnssecConfig property.

Line 62 is currently never hit because it is inside the if statement.


dave-pollock:
Failing tests appear to be unrelated to this change.

giorod3 (approved):

giorod3:
@dave-pollock thank you for this contribution!

PR #470, opened 2020-12-21
Fix: Google CorporateEmailsOnly Errors When Domain In IAM Policy

dave-pollock:
It's possible for a domain to be used instead of an email address in an IAM policy which breaks when the plugin splits the string using the `@` character.

This change will allow the plugin to continue without error if it encounters an IAM member without an `@` in the name.

### DIFF
--- a/plugins/google/iam/corporateEmailsOnly.js
+++ b/plugins/google/iam/corporateEmailsOnly.js
@@ -10,12 +10,12 @@ module.exports = {
     recommended_action: 'Ensure that no users are actively using their Gmail accounts to access GCP.',
     apis: ['projects:getIamPolicy'],
 
-    run: function(cache, settings, callback) {
+    run: function (cache, settings, callback) {
         var results = [];
         var source = {};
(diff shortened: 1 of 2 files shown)


plugins/google/iam/corporateEmailsOnly.js
```diff
@@ -10,12 +10,12 @@ module.exports = {
     recommended_action: 'Ensure that no users are actively using their Gmail accounts to access GCP.',
     apis: ['projects:getIamPolicy'],
 
-    run: function(cache, settings, callback) {
+    run: function (cache, settings, callback) {
```
AkhtarAmir, line 13: I think you should remove spaces around brackets here, line#18 and line#60 because we are not using spaces in these cases throughout the application so better keep it consistent.

dave-pollock: done

giorod3:
Thank you for the contribution! this has been merged internally

PR #472, opened 2020-12-22
feature/AKD-27: Added 'Lambda Env-vars Client Side Encrypted' plugin and test cases

AkhtarAmir:


plugins/aws/lambda/envVarsClientSideEncryption.js
```diff
@@ -0,0 +1,90 @@
...
+                    helpers.addResult(results, 0,
+                        'No environment variables found', region, resource);
+                    continue;
+                }
+
+                var unencryptedVars = [];
+                var encryptedVars = [];
+                for (var envVar of config.lambda_sensitive_env_vars) {
+                    if (lambdaFunc.Environment.Variables[envVar]) {
+                        if (lambdaFunc.Environment.Variables[envVar].length < 200) unencryptedVars.push(envVar);
```
dipsubha06, line 67: is it always the case that an encrypted var will have length > 200. what if we put ssh private key or some long string whose length is 200 and that is not encrypted. is there any other way to detect encryption.

dipsubha06 (changes requested):

plugins/aws/lambda/envVarsClientSideEncryption.js
AkhtarAmir: So Gio and I discussed and there is no way to know whether it is cipher text or value. So I changed the check to check for length to be exactly equal to 216 bytes as it is always that long.

dipsubha06 (approved):

plugins/aws/lambda/envVarsClientSideEncryption.js
```diff
@@ -0,0 +1,90 @@
...
+    more_info: 'AWS Lambda lets you encrypt environment variable values prior to sending them to Lambda. ' +
+        'Environment variables are often used to store sensitve information such as passwords. Such variable valuesshould be ' +
+        'encrypted for security best practices.',
+    link: 'https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html',
+    recommended_action: 'Encrypt environment variables that store sensitive information',
+    apis: ['Lambda:listFunctions'],
+    settings: {
+        lambda_sensitive_env_vars: {
+            name: 'Lambda Sensitive Environment Varibales',
+            description: 'A comma-delimited list of known lambda function Environment Varibales that should be encrypted',
```
giorod3, line 17:
```suggestion
            description: 'A comma-delimited list of known lambda function Environment Variables that should be encrypted',
```

giorod3 (approved):

PR #475, opened 2020-12-25
feature/spec: Added test cases for EC2 plugins

AkhtarAmir:


plugins/aws/ec2/openDocker.spec.js
```diff
@@ -0,0 +1,139 @@
...
+                "FromPort": 0,
+                "IpProtocol": "tcp",
+                "IpRanges": [
+                    {
+                        "CidrIp": "0.0.0.0/0"
+                    }
+                ],
+                "Ipv6Ranges": [],
+                "PrefixListIds": [],
+                "ToPort": 65535,
```
dipsubha06, line 28: Docker port is 'tcp': [2375, 2376]. But here in the test cases those are not open.

plugins/aws/ec2/openFTP.spec.js
```diff
@@ -0,0 +1,139 @@
...
+                "FromPort": 0,
+                "IpProtocol": "tcp",
+                "IpRanges": [
+                    {
+                        "CidrIp": "0.0.0.0/0"
+                    }
+                ],
+                "Ipv6Ranges": [],
+                "PrefixListIds": [],
+                "ToPort": 65535,
```
dipsubha06, line 28: same here FTP port is 20 and 21. But in failing case we dont have test cases with these two open

dipsubha06 (changes requested):
In open port test cases please check if the failing test case is having the concerned ports open.

plugins/aws/ec2/openDocker.spec.js
AkhtarAmir: I am sending all ports open which are 0-65535 so technically that covers 'tcp': [2375, 2376] as well.

plugins/aws/ec2/openFTP.spec.js
AkhtarAmir: I am sending all ports open which are 0-65535 so technically that covers 'tcp': [20, 21] as well.

giorod3 (approved):

PR #514, opened 2021-01-22
Check Kubernetes Cluster scopes independently

stevekrouse:
I may be missing something, but it seems like there's a bug how it's currently written: if any Cluster has a bad scope, all subsequent Clusters will also be marked as out-of-scope because this variable isn't reset on each run.

### DIFF
--- a/plugins/google/kubernetes/clusterLeastPrivilege.js
+++ b/plugins/google/kubernetes/clusterLeastPrivilege.js
@@ -43,9 +43,8 @@ module.exports = {
                 'https://www.googleapis.com/auth/trace.append'
             ];
 
-            let otherScope = false;
-
             clusters.data.forEach(cluster => {
+                let otherScope = false;
                 if (cluster.nodeConfig &&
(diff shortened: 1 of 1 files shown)


stevekrouse:
I think the code can be simplified even further, but I'll leave that in this comment and defer to you whether you want to put it in the PR. Instead of:

```js
                      cluster.nodeConfig.oauthScopes.forEach((oneScope) => {
                        let sameExist= false;
                        for (let i = 0; i < minimalAccess.length; i++) {
                            if (oneScope == minimalAccess[i]) {
                                sameExist = true;
                            }
                        }
                        if (sameExist == false) {
                            otherScope = true;
                        }
                    });
                }
                if (otherScope == true) {
```

You could have:

```js
                if (cluster.nodeConfig.oauthScopes.some(scope => !minimalAccess.includes(scope))) {
```

giorod3 (approved):

giorod3:
@stevekrouse Thank you for this contribution!

stevekrouse:
Pleasure :)

PR #515, opened 2021-01-24
feature/AKD-35: Added AWS S3 'Bucket Transfer Acceleration' plugin and test cases

AkhtarAmir:


giorod3 (changes requested):
remove package-lock and reset package

giorod3 (approved):

PR #534, opened 2021-02-08
feature/akd-23: Added AWS 'Glue CloudWatch Logs Encryption' plugin and test cases

AkhtarAmir:


plugins/aws/glue/glueCloudwatchLogsEncrypted.js
```diff
@@ -0,0 +1,65 @@
...
+
+            getSecurityConfigurations.data.forEach(configuration => {
+                if (!configuration.Name) return;
+
+                var resource = `arn:${awsOrGov}:glue:${region}:${accountId}:/securityConfiguration/${configuration.Name}`;
+
+                if(configuration.EncryptionConfiguration &&
+                    configuration.EncryptionConfiguration.CloudWatchEncryption &&
+                    configuration.EncryptionConfiguration.CloudWatchEncryption.CloudWatchEncryptionMode &&
+                    configuration.EncryptionConfiguration.CloudWatchEncryption.CloudWatchEncryptionMode === 'SSE-KMS') {
```
dipsubha06, line 48: is sse-kms the only option for encryption here. what about CMK. if that is also allowed we should check for that as well.

AkhtarAmir: Yes it has only two options: Disabled or SSE-KMS

dipsubha06: Okay. then we are good here.

dipsubha06 (approved):

giorod3 (approved):

giorod3 (approved):

PR #539, opened 2021-02-12
modification: Modified to fix false FAIL results

AkhtarAmir:


plugins/aws/route53/domainExpiry.js
```diff
@@ -31,17 +31,15 @@ module.exports = {
             return callback(null, results, source);
         }
 
-        for (var i in listDomains.data) {
-            var domain = listDomains.data[i];
-
+        for (var domain of listDomains.data) {
             if (domain.Expiry) {
-                var difference = helpers.daysAgo(domain.Expiry);
+                var difference = Math.round((new Date(domain.Expiry).getTime() - new Date().getTime())/(24*60*60*1000));
```
dipsubha06, line 36: Cant we use the daysAgo? the number of days only changed to be 35. so cant we get the value from daysAgo()

AkhtarAmir: daysAgo function is doing Math.abs on result. So it converts -ve days to +ve and domain already expired message will never be shown

giorod3 (approved):

PR #542, opened 2021-02-12
feature/83: Added 'Classic Load Balancer In Use' plugin and test cases

AkhtarAmir:


plugins/aws/elb/classicELBInUse.js
```diff
@@ -0,0 +1,73 @@
...
+                    helpers.addResult(results, 0,
+                        `ELB "${lb.LoadBalancerName}" is not using any listeners`,
+                        region, elbArn);
+                    return cb();
+                }
+
+                for (var listener of lb.ListenerDescriptions){
+                    if (lb.Instances && lb.Instances.length && listener.Listener && ( listener.Listener.Protocol === 'HTTP' || listener.Listener.Protocol === 'HTTPS')) {
+                        classicLBFound = true;
+                        helpers.addResult(results, 2,
```
giorod3, line 53: move this result message to outside the for loop. this could also be a filter if need be.

giorod3 (approved):

giorod3 (approved):

PR #563, opened 2021-03-01
ensure pluginDone is called under all circumstances

### LINKED ISSUE #552: Nothing Collect AWS, only compliance
The standard AWS scanners are coming in empty, only the compliance scans like PCI, HIPAA, CIS present information.

### LINKED ISSUE #557: --json=file.json creates empty json file
when running tool agains AWS the csv file is fully populated, however json file is created but empty ( length is 0 ). No errors shown.

shamelesscookie:
Fixes #552, Fixes #557 

The implementation of [async.mapValuesLimit](https://caolan.github.io/async/v3/mapValuesLimit.js.html) in `engine.collector` fails to resolve all of the asynchronous functions. This prevents the callback function from executing, which means `outputHandler.close()` never gets called. That means output is never written to the console, nor saved to JSON.

If we move the call to `pluginDone` out of the `if` statement, it is called under all circumstances, ensuring we get proper output.

### DIFF
--- a/engine.js
+++ b/engine.js
@@ -190,7 +190,6 @@ var engine = function(cloudConfig, settings) {
                     }
 
                 }
-                setTimeout(function() { pluginDone(err, maximumStatus); }, 0);
             };
 
             if (plugin.gsl) {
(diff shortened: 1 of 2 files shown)


shamelesscookie:
I think the test case for `accessKeysRotated` is prone to false negative. The test spec is initialized with:

```
warnDate.setMonth(warnDate.getMonth() - 3);
```

and the plugin makes its determination based on > (not >=):

```
if (helpers.daysAgo(userCreationTime) > config.access_keys_rotated_warn
```

which means the warnDate in the test case won't actually generate a warning since it's exactly 90 days old. If you change the test spec to, e.g. 

```
warnDate.setMonth(warnDate.getMonth() - 4);
```

the test passes

jakobheine:
@shamelesscookie you are the real MVP

PR #568, opened 2021-03-04
HOTFIX/SAAS-1247: Fixed false unknown in case of govcloud

AkhtarAmir:


plugins/aws/iam/rootMfaEnabled.js
```diff
@@ -52,7 +52,9 @@ module.exports = {
             }
         }
 
-        if (!found) {
+        if (settings.govcloud && !found) {
```
dipsubha06, line 55: Do we need to check for !found? instead we can check only for govcloud and that is enough is not it?

AkhtarAmir: Yes that is enough. I just wanted to be sure that it does not produce two results. What do you think?

dipsubha06 (approved):

giorod3 (approved):

PR #589, opened 2021-03-14
spec: Added test case file for Enforce MySQL SSL Connection plugin

ali-imran7:


plugins/azure/mysqlserver/enforceMySQLSSLConnection.spec.js
```diff
@@ -0,0 +1,103 @@
...
+    },
+    {
+        sku: [Object],
+        location: 'eastus',
+        tags: {},
+        id: '/subscriptions/123/resourceGroups/aqua-resource-group/providers/Microsoft.DBforMySQL/servers/test-server',
+        name: 'test-server',
+        type: 'Microsoft.DBforMySQL/servers',
+        sslEnforcement: 'Disabled',
+        userVisibleState: 'Ready',
```
AkhtarAmir, line 27: Use double quotes notation to define objects to keep consistent
#586

AkhtarAmir (changes requested):

plugins/azure/mysqlserver/enforceMySQLSSLConnection.spec.js
ali-imran7: Done as suggested.
@AkhtarAmir

giorod3 (approved):

PR #630, opened 2021-04-01
remediation: Azure App Service Idnetity Enabled plugin remediated

AkhtarAmir:


plugins/azure/appservice/identityEnabled.js
```diff
@@ -48,5 +54,50 @@ module.exports = {
...
+        // for logging purposes
+        var webAppNameArr = resource.split('/');
+        var webAppName = webAppNameArr[webAppNameArr.length - 1];
+
+        // create the params necessary for the remediation
+        if (settings.region) {
+            var body = {
+                'location': settings.region,
+                'identity': {
+                    'type': 'SystemAssigned'
```
dipsubha06, line 76: What if user wants to give user assigned managed identity here? can we take that as input and if no input we can fall back to system assigned one.

dipsubha06 (changes requested):

giorod3 (approved):

PR #635, opened 2021-04-04
feature/AKD-148: Added Azure 'No Unattached Disk Volumes' plugin and test cases

ali-imran7:


dipsubha06 (approved):
While the code changes looks fine, but I have this point to highlight. Disks are basically the place to hold data. and charges for disks are lower than a running vm. so sometimes i have seen customers keep unattached disks with data so that when needed they use them by attaching to a vm.

plugins/azure/virtualmachines/noUnattachedDisks.js
```diff
@@ -0,0 +1,48 @@
...
+                return rcb();
+            }
+            if (!disks.data.length) {
+                helpers.addResult(results, 0, 'No existing disk volumes found', location);
+                return rcb();
+            }
+
+            disks.data.forEach(disk => {
+                if (disk.diskState && disk.diskState.toLowerCase() === 'attached') {
+                    helpers.addResult(results, 0, 'Disk volume is attached to a virtual machine.', location, disk.id);
```
AkhtarAmir, line 36:
```suggestion
                    helpers.addResult(results, 0, 'Disk volume is attached to a virtual machine', location, disk.id);
```

plugins/azure/virtualmachines/noUnattachedDisks.js
```diff
@@ -0,0 +1,48 @@
...
+                } else {
+                    helpers.addResult(results, 2, 'Disk volume is not attached to a virtual machine.', location, disk.id);
```
AkhtarAmir, line 38:
```suggestion
                    helpers.addResult(results, 2, 'Disk volume is not attached to a virtual machine', location, disk.id);
```

AkhtarAmir (approved):

ali-imran7:
> While the code changes looks fine, but I have this point to highlight. Disks are basically the place to hold data. and charges for disks are lower than a running vm. so sometimes i have seen customers keep unattached disks with data so that when needed they use them by attaching to a vm.

@dipsubha06 In that case, don't they have option to suppress the plugin?

giorod3 (approved):

PR #648, opened 2021-04-09
feature-iam-policies-present: Added AWS IAM Policies Present plugin and test cases

AkhtarAmir:


plugins/aws/iam/iamPoliciesPresent.js
```diff
@@ -0,0 +1,105 @@
...
+
+            if (listRolePolicies.data && listRolePolicies.data.PolicyNames) attachedPolicies = attachedPolicies.concat(listRolePolicies.data.PolicyNames);
+
+            for (let policy of config.iam_required_policy_names) {
+                if (!attachedPolicies.includes(policy)) difference.push(policy);
+            }
+
+            if (difference.length) {
+                helpers.addResult(results, 2,
+                    `IAM role does not have these required policies attached: ${difference.join(', ')}`, region, role.RoleName);
```
giorod3, line 94:
```suggestion
                    `IAM role does not have these required policies attached: ${difference.join(', ')}`, region, role.Arn);
```

plugins/aws/iam/iamPoliciesPresent.js
```diff
@@ -0,0 +1,105 @@
...
+            } else {
+                helpers.addResult(results, 0,
+                    'IAM role has all required policies attached', region, role.RoleName);
```
giorod3, line 97:
```suggestion
                    'IAM role has all required policies attached', region, role.Arn);
```

PR #664, opened 2021-04-22
Added GCP SQL plugins for database flags.

umermehmood:


plugins/google/sql/postgresqlLogMinDuration.js
```diff
@@ -0,0 +1,68 @@
...
+                                                                        flag.value && flag.value == '-1');
+
+                        if (found) {
+                            helpers.addResult(results, 0, 
+                                'SQL instance have log_min_duration_statement flag enabled', region, sqlInstance.name);
+                        } else {
+                            helpers.addResult(results, 2,
+                                'SQL instance does not have log_min_duration_statement flag enabled', region, sqlInstance.name);
+                        }
+                } else if (sqlInstance.instanceType == "READ_REPLICA_INSTANCE") {
```
dipsubha06, line 55: Is this else if needed. if so can we have some handling in that case

plugins/google/sql/postgresqlLogMinDuration.js
```diff
@@ -0,0 +1,68 @@
...
+
+                if (sqlInstance.instanceType != "READ_REPLICA_INSTANCE" &&
+                    sqlInstance.settings &&
+                    sqlInstance.settings.databaseFlags &&
+                    sqlInstance.settings.databaseFlags.length) {
+                        let found = sqlInstance.settings.databaseFlags.find(flag => flag.name && flag.name == 'log_min_duration_statement' &&
+                                                                        flag.value && flag.value == '-1');
+
+                        if (found) {
+                            helpers.addResult(results, 0,
```
dipsubha06, line 49: from the description of the plugin it is said that the flag should not be enabled. then this should be a failing case if the flag is enabled,right?

dipsubha06 (changes requested):

plugins/google/sql/postgresqlLogTempFiles.js
```diff
@@ -0,0 +1,68 @@
+var async = require('async');
+var helpers = require('../../../helpers/google');
+
+module.exports = {
+    title: 'PostgreSQL Log Temp Files Flag Enabled',
```
AkhtarAmir, line 5:
```suggestion
    title: 'PostgreSQL Log Temp Files',
```

AkhtarAmir (changes requested):

plugins/google/sql/postgresqlLogMinDuration.js
```diff
@@ -0,0 +1,68 @@
+var async = require('async');
+var helpers = require('../../../helpers/google');
+
+module.exports = {
+    title: 'PostgreSQL Log Min Duration Statement',
+    category: 'SQL',
+    description: 'Ensures SQL instances for PostgreSQL type have log min duration statement flag disabled.',
+    more_info: 'SQL instance for PostgreSQL databases provides log_min_duration_statement flag. It is used to log the duration of every completed statement. We should always disable it as there can be sensitive information as well that should not be recorded in the logs.',
```
giorod3, line 8:
```suggestion
    more_info: 'SQL instance for PostgreSQL databases provides log_min_duration_statement flag. It is used to log the duration of every completed statement. This should always be disabled as there can be sensitive information as well that should not be recorded in the logs.',
```

plugins/google/sql/postgresqlLogMinDuration.js
```diff
@@ -0,0 +1,68 @@
...
+                        if (found) {
+                            helpers.addResult(results, 0, 
+                                'SQL instance does not have log_min_duration_statement flag enabled', region, sqlInstance.name);
+                        } else {
+                            helpers.addResult(results, 2,
+                                'SQL instance have log_min_duration_statement flag enabled', region, sqlInstance.name);
+                        }
+                } else if (sqlInstance.instanceType == "READ_REPLICA_INSTANCE") {
+                } else {
+                    helpers.addResult(results, 2,
```
giorod3, line 57: this should be a pass if there are no database flags

giorod3 (changes requested):
two comments

giorod3 (approved):

PR #687, opened 2021-05-04
SQL plugin and spec file added for checking public access.

umermehmood:


plugins/google/sql/sqlPublicAccessDisabled.js
```diff
@@ -0,0 +1,57 @@
...
+                       
+                        if (found) {
+                            helpers.addResult(results, 2, 
+                                'SQL instance have public access enabled', region, sqlInstance.name);
+                        } else {
+                            helpers.addResult(results, 0,
+                                'SQL instance does not have public access enabled', region, sqlInstance.name);
+                        }
+                } else if (sqlInstance.instanceType == "READ_REPLICA_INSTANCE") {
+                }
```
giorod3, line 48: we should never have an empty else if, also we should always produce a result per resource

umermehmood: I saw this scenario in multiple plugins e.g. (dbPubliclyAccessible). And added this as a standard as we are not validating the read replicas. However, I have removed the empty else if from here. And the result is generated per resource.

plugins/google/sql/sqlPublicAccessDisabled.js
```diff
@@ -0,0 +1,56 @@
+var async = require('async');
+var helpers = require('../../../helpers/google');
+
+module.exports = {
+    title: 'SQL Public Access Disabled',
+    category: 'SQL',
+    description: 'Ensures SQL instances does not have public access enabled.',
+    more_info: 'Public access can cause security issues. So it should always be disabled.',
+    link: 'https://www.readitquik.com/news/cloud-3/google-cloud-platform-introduces-private-networking-for-cloud-sql/#',
```
AkhtarAmir, line 9: Use official doc

plugins/google/sql/sqlPublicAccessDisabled.js
```diff
@@ -0,0 +1,56 @@
+var async = require('async');
+var helpers = require('../../../helpers/google');
+
+module.exports = {
+    title: 'SQL Public Access Disabled',
+    category: 'SQL',
+    description: 'Ensures SQL instances does not have public access enabled.',
```
AkhtarAmir, line 7:
```suggestion
    description: 'Ensure that SQL instances are using private IPs instead of public IPs.',
```

plugins/google/sql/sqlPublicAccessDisabled.js
```diff
@@ -0,0 +1,56 @@
+var async = require('async');
+var helpers = require('../../../helpers/google');
+
+module.exports = {
+    title: 'SQL Public Access Disabled',
```
AkhtarAmir, line 5:
```suggestion
    title: 'SQL No Public IPs',
```

plugins/google/sql/sqlPublicAccessDisabled.js
```diff
@@ -0,0 +1,56 @@
...
+    category: 'SQL',
+    description: 'Ensures SQL instances does not have public access enabled.',
+    more_info: 'Public access can cause security issues. So it should always be disabled.',
```
AkhtarAmir, line 8:
```suggestion
    more_info: 'Cloud SQL databases should always use private IP addresses which provide improved network security and lower latency.',
```

plugins/google/sql/sqlPublicAccessDisabled.js
```diff
@@ -0,0 +1,56 @@
...
+    link: 'https://www.readitquik.com/news/cloud-3/google-cloud-platform-introduces-private-networking-for-cloud-sql/#',
+    recommended_action: 'Ensure that public access is disabled for all SQL instances.',
```
AkhtarAmir, line 10:
```suggestion
    recommended_action: 'Make sure that SQL databases IP addresses setting does not have IP address of PRIMARY type',
```

AkhtarAmir (changes requested):

plugins/google/sql/sqlPublicAccessDisabled.js
AkhtarAmir: I have moved the check to the top

AkhtarAmir (approved):

giorod3 (approved):

PR #692, opened 2021-05-09
GCP storage plugin added for uniform level access with spec file.

umermehmood:


plugins/google/storage/bucketUniformAccess.js
```diff
@@ -0,0 +1,54 @@
...
+                    if (bucket.iamConfiguration &&
+                        bucket.iamConfiguration.uniformBucketLevelAccess && 
+                        bucket.iamConfiguration.uniformBucketLevelAccess.enabled) {
+                        helpers.addResult(results, 0, 'Bucket has uniform bucket level access enabled', region, bucket.id);
+                    } else {
+                        helpers.addResult(results, 2, 'Bucket does not have uniform bucket level access enabled', region, bucket.id);
+                    }
+                } else {
+                    helpers.addResult(results, 0, 'No storage buckets found', region);
+                    return
```
giorod3, line 44: you don't need this return its the last statement in the loop

plugins/google/storage/bucketUniformAccess.js
```diff
@@ -0,0 +1,54 @@
...
+                if (bucket.id) {
+                    if (bucket.iamConfiguration &&
+                        bucket.iamConfiguration.uniformBucketLevelAccess && 
+                        bucket.iamConfiguration.uniformBucketLevelAccess.enabled) {
+                        helpers.addResult(results, 0, 'Bucket has uniform bucket level access enabled', region, bucket.id);
+                    } else {
+                        helpers.addResult(results, 2, 'Bucket does not have uniform bucket level access enabled', region, bucket.id);
+                    }
+                } else {
+                    helpers.addResult(results, 0, 'No storage buckets found', region);
```
giorod3, line 43: why would the else of bucket.id mean no storage buckets found?

umermehmood: API returns the array with one object even if there are no buckets found. This is to handle that scenario.

plugins/google/storage/bucketUniformAccess.js
```diff
@@ -0,0 +1,53 @@
+var async = require('async');
+var helpers = require('../../../helpers/google');
+
+module.exports = {
+    title: 'Bucket Uniform Level Access',
+    category: 'Storage',
+    description: 'Ensures uniform level access is enabled on storage buckets',
+    more_info: 'Uniform level access for buckets can be used for managing access in a simple way. It enables us to use other security features like IAM conditions',
+    link: 'https://cloud.google.com/storage/docs/uniform-bucket-level-access#should-you-use',
+    recommended_action: 'Uniform level access should be enabled for the bucket, it provides simple ways to manage the access. Also enables us to use other security features like domain restricted sharing and IAM conditions',
```
AkhtarAmir: This should be a concise recommendation for the action to be taken from user's end.

AkhtarAmir (changes requested):

plugins/google/storage/bucketUniformAccess.js
AkhtarAmir: I have changes the check

AkhtarAmir (approved):

giorod3 (approved):

PR #702, opened 2021-05-13
Alibaba ram user plugin added for inactive user disabled

umermehmood:


plugins/alibaba/ram/inactiveUserDisabled.js
```diff
@@ -0,0 +1,65 @@
...
+                if (getUserProfile.data.LoginProfile) {
+                    helpers.addResult(results, 2,
+                        'RAM user inactive for 90 or more days is enabled ', region, resource);
+                } else {
+                    helpers.addResult(results, 0,
+                        'RAM user inactive for 90 or more days is not enabled', region, resource);
+                }
+            } else {
+                helpers.addResult(results, 0,
+                    'RAM user inactive for 90 or more days is not enabled', region, resource);
```
dipsubha06, line 59: here it is checked that if the diffdays is not greater or equal 90, but we are not checking the login profile, But in message we are saying user is not enabled. as our check is only for user with days limit >=90 here i think we should not have a else at all as we are not interested there, or we need to check the user profile as well before saying its not enabled. what if some user has kept enabled but last access is less than 90. it will give result as  as not enabled.

dipsubha06 (changes requested):

AkhtarAmir (changes requested):

giorod3 (approved):

PR #704, opened 2021-05-14
Feature/166 - GCP computer instance default service account plugin with test cases

umermehmood:


plugins/google/compute/insDefaultServiceAccount.js
```diff
@@ -0,0 +1,88 @@
...
+                helpers.addResult(results, 0, 'No instances found in the region' , region);
+            } if (defaultServiceAccountInstances.length) {
+                var myInstanceStr = defaultServiceAccountInstances.join(", ");
+                helpers.addResult(results, 2,
+                    `Default service account is used for following instances: ${myInstanceStr}`, region);
+            } if (validServiceAccountInstances.length) {
+                var myInstanceStr = validServiceAccountInstances.join(", ");
+                helpers.addResult(results, 0,
+                    `Default service account is not used for following instances: ${myInstanceStr}`, region);
+            }
```
giorod3, line 81: are these supposed to be else if's?

umermehmood: There can be multiple instances in one zone. And multiple instances can have different settings. That's why multiple conditions can be fulfilled at the same iteration.

giorod3 (approved):

PR #715, opened 2021-05-16
feature/AKD-144: Added Azure 'Scale Sets Health Monitoring Enabled' plugin and test cases

ali-imran7:


plugins/azure/virtualmachines/scaleSetHealthMonitoring.js
```diff
@@ -0,0 +1,61 @@
...
+
+            async.each(virtualMachineScaleSets.data, (virtualMachineScaleSet, scb) => {
+                const scaleSetExtensions = virtualMachineScaleSet.virtualMachineProfile && virtualMachineScaleSet.virtualMachineProfile.extensionProfile &&
+                virtualMachineScaleSet.virtualMachineProfile.extensionProfile.extensions 
+                    ? virtualMachineScaleSet.virtualMachineProfile.extensionProfile.extensions 
+                    : [];
+
+                const healthMonitoring = scaleSetExtensions.length 
+                    ? scaleSetExtensions.some((extension) => (extension.name === 'healthRepairExtension' || extension.type === 'ApplicationHealthLinux')) 
+                    : false;
```
AkhtarAmir, line 43: I think find is more appropriate for this check

AkhtarAmir (changes requested):

plugins/azure/virtualmachines/scaleSetHealthMonitoring.js
ali-imran7: we don't have to check whats in the extension so true or false can work,

giorod3 (approved):

PR #761, opened 2021-06-15
feature/AKD-173: Added Google Network DNS Logging Enabled plugin and test cases

AkhtarAmir:


PickluPaul:
@AkhtarAmir  I am trying to exctract the selflink for the GCP resources from into the output results, however, I am not able to extract it. Any pointers on how to include this self link into the output.

giorod3 (approved):

PR #781, opened 2021-06-24
feature/AKD-283: Plugin to check bucket transfer acceleration and unit tests

UmairAjmal:


exports.js
```diff
@@ -796,6 +796,7 @@ module.exports = {
       
         'bucketLoggingEnabled'          : require(__dirname + '/plugins/alibaba/oss/bucketLoggingEnabled.js'),
         'ossBucketPrivate'              : require(__dirname + '/plugins/alibaba/oss/ossBucketPrivate.js'),
+        'ossBucketTransferAcceleration' : require(__dirname + '/plugins/alibaba/oss/ossBucketTransferAcceleration.js'),
```
AkhtarAmir, line 799:
```suggestion
        'BucketTransferAcceleration'    : require(__dirname + '/plugins/alibaba/oss/BucketTransferAcceleration.js'),
```

AkhtarAmir (changes requested):

exports.js
UmairAjmal: According to the naming convention, plugin names do not start with a capital letter. 'bucketTransferAcceleration' pulgin already exists. Are you sure to rename as you suggested?

giorod3 (approved):

PR #785, opened 2021-06-30
feature/AKD-180: Added 'Minimum TLS Version' plugin and test cases

ali-imran7:


plugins/azure/redisCache/minimumTlsVersion.js
```diff
@@ -0,0 +1,48 @@
...
+
+            if (!caches.data.length) {
+                helpers.addResult(results, 0, 'No Redis Caches found', location);
+                return rcb();
+            }
+
+            for (let cache of caches.data) {
+                if (cache.minimumTlsVersion && (cache.minimumTlsVersion === '1.0' || cache.minimumTlsVersion === '1.1')) {
+                    helpers.addResult(results, 2, 'Redis Cache is not using the latest TLS Version', location, cache.id);
+                } else {
```
dipsubha06, line 37: should we have a else if case for !cache.minimumTlsVersion and report an error saying unable to get the tls version

dipsubha06 (changes requested):

plugins/azure/redisCache/minimumTlsVersion.js
AkhtarAmir: @ali-imran7 verify this. I think if we don't specify any version, then it 1.0 by default and I guess it does not return version in that case

ali-imran7: Yeah @dipsubha06 I agree. we should have one. 

@AkhtarAmir you are right. API does not return anything in case of the default version set. So, I added a case to generate an error with the message "Redis Cache is using the default TLS Version" .

giorod3 (approved):

PR #792, opened 2021-07-05
modification/SAAS-1956: Added new regions

AkhtarAmir:


AkhtarAmir:
Added new regions for Instance Max Count plugin for instance limit settings

giorod3 (approved):

PR #794, opened 2021-07-06
fix/blobContainersPrivateAccess : Added check for public access flag at storage account level

ali-imran7:


plugins/azure/blobservice/blobContainersPrivateAccess.js
```diff
@@ -42,25 +42,29 @@ module.exports = {
...
-                    cache, source, ['blobContainers', 'list', location, storageAccount.id]
-                );
-
-                if (!blobContainers || blobContainers.err || !blobContainers.data) {
-                    helpers.addResult(results, 3,
-                        'Unable to query Blob Containers: ' + helpers.addError(blobContainers),
-                        location, storageAccount.id);
-                } else if (!blobContainers.data.length) {
-                    helpers.addResult(results, 0, 'Storage Account does not contain blob containers', location, storageAccount.id);
+                if (storageAccount.allowBlobPublicAccess === false) {
```
AkhtarAmir, line 45:
```suggestion
                if (!storageAccount.allowBlobPublicAccess) {
```

AkhtarAmir (approved):

AkhtarAmir:
Added a check on storage account level if storage account has blocked public access for blob then it should not check for block container

giorod3 (approved):

PR #801, opened 2021-07-13
SAAS - 1828 : Added Default Mapping for security center policy assignments

ali-imran7:


AkhtarAmir:
Modified to create a map for default values for security center monitor policies so if they are being returned in the response, the plugin should check for their default value and display result accordingly.
Also some of the policies are deprecated so adding a new check for those to pass.

giorod3 (approved):

PR #872, opened 2021-08-29
SAAS-2469: Google Compute Engine Persistent Disks Auto Delete

mehakseedat63:


plugins/google/compute/persistentDisksAutoDelete.js
```diff
@@ -0,0 +1,82 @@
...
+
+        let projects = helpers.addSource(cache, source,
+            ['projects', 'get', 'global']);
+
+        if (!projects || projects.err || !projects.data || !projects.data.length) {
+            helpers.addResult(results, 3,
+                'Unable to query for projects: ' + helpers.addError(projects), 'global', null, null, (projects) ? projects.err : null);
+            return callback(null, results, source);
+        }
+
```
AkhtarAmir, line 26: We should user disks.list API and resource should be the actual disk instead of VM instance. I see that disks.list does not return autoDelete flag so we can first call VM instances API and get all disks which have autoDelete enabled as we are doing right now, and then call disks.list and check each disk against that list. Also we need to have a setting for disk_result_limit as in CSEK Encryption Enabled plugin.

AkhtarAmir (changes requested):

plugins/google/compute/persistentDisksAutoDelete.js
```diff
@@ -0,0 +1,124 @@
...
+    more_info: 'When auto-delete is enabled, the attached persistent disk are deleted with VM instance deletion. In cloud environments, you might want to keep the attached persistent disks even when the associated VM instance is deleted.',
+    link: 'https://cloud.google.com/compute/docs/disks',
+    recommended_action: 'Ensure that auto-delete is disabled for all disks associated with your VM instances.',
+    apis: ['disks:list', 'instances:compute:list', 'projects:get'],
+    settings: {
+        disk_result_limit: {
+            name: 'Persistent Disks Auto Delete Result Limit',
+            description: 'If the number of results is greater than this value, combine them into one result',
+            regex: '^[0-9]*$',
+            default: '20',
```
AkhtarAmir, line 17:
```suggestion
            default: '20'
```

plugins/google/compute/persistentDisksAutoDelete.js
```diff
@@ -0,0 +1,124 @@
...
+                                }
+                            });
+                        }
+                    });
+                }
+
+                disks.data.forEach(disk => {
+                    if (!disk.id || !disk.selfLink || !disk.creationTimestamp) return;
+
+                    if (disk.autoDelete && !(badDisks.find(badDisk => badDisk.source == disk.selfLink))) {
```
AkhtarAmir, line 81: Change this

AkhtarAmir (changes requested):

AkhtarAmir (approved):

Issue #960, opened 2021-11-09
GCP openAllPorts error


amulroney-del:
Hello All,

I am working on validating some policies for an application and have been getting an error on the GCP openAllPorts policy. Here is the output:  
  
'''

$ node ./index.js --config ./config.js --cloud google --json=results.json --collection=collection.json --plugin openAllPorts

   _____ _                 _  _____       _       _ _
  / ____| |               | |/ ____|     | |     (_) |
 | |    | | ___  _   _  __| | (___  _ __ | | ___  _| |_
 | |    | |/ _ \| | | |/ _` |\___ \| '_ \| |/ _ \| | __|
 | |____| | (_) | |_| | (_| |____) | |_) | | (_) | | |_
  \_____|_|\___/ \__,_|\__,_|_____/| .__/|_|\___/|_|\__|
                                   | |
                                   |_|

  CloudSploit by Aqua Security, Ltd.
  Cloud security auditing for AWS, Azure, GCP, Oracle, and GitHub

INFO: Using CloudSploit config file: ./config.js
INFO: Skipping AWS pagination mode
INFO: Testing plugin: Open All Ports
INFO: Determining API calls to make...
INFO: Found 1 API calls to make for google plugins
INFO: Collecting metadata. This may take several minutes...
INFO: Metadata collection complete. Analyzing...
INFO: Analysis complete. Scan report to follow...
INFO: Collection file written to collection.json
INFO: JSON file written to results.json

  ┌──────────────────┬───────────────┬─────────────────────────────────────────────┬─────────┬───────┬──────────────────┬──────────────────────────────────────────────────┐
  │     Category     │    Plugin     │                 Description                 │ Resourc │ Regio │      Status      │                     Message                      │
  │                  │               │                                             │    e    │   n   │                  │                                                  │
  ├──────────────────┼───────────────┼─────────────────────────────────────────────┼─────────┼───────┼──────────────────┼──────────────────────────────────────────────────┤
  │ VPC Network      │ Open All      │ Determines if all ports are open to the     │ N/A     │ globa │ UNKNOWN          │ Unable to query for projects: Unable to obtain   │
  │                  │ Ports         │ public                                      │         │ l     │                  │ data                                             │
  ├──────────────────┼───────────────┼─────────────────────────────────────────────┼─────────┼───────┼──────────────────┼──────────────────────────────────────────────────┤
[...]

PR #1042, opened 2021-12-16
960: Allow Multi-Valued Conditions in IAM Policies

chrisoverzero:
The code here assumed that any condition value must be a string, but
condition values may also be arrays. The regex tests succeed in many
cases because the array stringifies to a comma-delimited list of its
elements, but this _also_ means that only the first element is
checked, effectively. When the array-of-arrays is passed back to, say,
`sqsCrossAccount`, the `includes` test _works_, but not in the way
which the author intended. This causes false negatives.

Resolves: #960


chrisoverzero:
Viewing this diff while ignoring whitespace differences will show how small this change truly is.

chrisoverzero:
Looks as though this was fixed for SQS in #991. I don't know if there's any interest in what I think is a more complete solution. If not, please do feel free to close this.

PR #1048, opened 2021-12-26
SLK-43687: Added setting to only WARN federated users

AkhtarAmir:


plugins/oracle/identity/usersMfaEnabled.js
```diff
@@ -17,10 +17,22 @@ module.exports = {
         pci: 'PCI requires MFA for all access to cardholder environments. ' +
             'Create an MFA key for user accounts.'
     },
+    setting: {
+        warn_federated_users: {
+            name: 'Warn Federated Users',
+            description: 'Give a WARN instead of FAIL result for federated users',
+            regex: '^(true|false)$',
+            default: 'true'
```
AkhtarAmir, line 25: this should be false

dipsubha06:
```suggestion
            default: 'false'
```

dipsubha06 (approved):

PR #1119, opened 2022-01-26
fix-broken-node-bin-links

pasanchamikara:
Fixed the issue in Dockerfile, which was observed while building. The Dockerfile, wasn't changing building the link references of the node app. So needed to rebuild them by adding a `npm link`


gonda-praetorian (approved):
Confirmed that this fixes Dockerfile.

AkhtarAmir (approved):

PR #1164, opened 2022-03-01
SLK-47764: Added Some Changes to SQS Encryption Plugin

sadeed12345:


plugins/aws/sqs/sqsEncryptionEnabled.js
```diff
@@ -43,13 +43,13 @@ module.exports = {
...
 
-            if (listQueues.err || !listQueues.data) {
+            if (listQueues.err) {
                 helpers.addResult(results, 3,
                     'Unable to query SQS queues: ' + helpers.addError(listQueues), region);
                 return rcb();
             }
 
-            if (!listQueues.data.length) {
+            if (!listQueues.data.length || !listQueues.data) {
```
AkhtarAmir, line 52:
```suggestion
            if (!listQueues.data || !listQueues.data.length) {
```

PR #1261, opened 2022-05-19
Fix YAML for GCP Role

mdtro:
The YAML used to define the GCP role had a small syntax error. See issue #909.


mdtro:
Cancelling. I see that #976 covers this. :)

PR #1455, opened 2022-11-07
Feature/cloudsploit gcp new rules p1

mcplima:
New GCP Rules - Cloudsploit - Part 1 (all rules with DOCs on Zanshin)
2 API rules
2 Compute rules
1 kms rule
2 sql rules
10 network rules

Updated collector.js with new collectors
Updated export.js with the new rules
Updated regions.js with the scopes for the new collectors

All these rules where documented previously on Zanshin. They refer to an update done to Cloudsploit 6 months ago.


alphadev4:
hi @mcplima can you please resolve the conflicts?

PR #1525, opened 2023-01-28
SAAS-13294: GCP - API Key Active Services Plugin

mehakseedat63:


plugins/google/api/apiKeyActiveServices.js
```diff
@@ -0,0 +1,69 @@
+var helpers = require('../../../helpers/google');
+
+module.exports = {
+    title: 'API Key Active Services Only',
+    category: 'API',
+    domain: 'Identity and Access Management',
+    description: 'Ensure API Keys only exist for active services.',
+    more_info: 'API Keys should only be used for services in cases where other authentication methods are unavailable. Unused keys with their permissions in tact may still exist within a project. Keys are insecure because they can be viewed publicly, such as from within a browser, or they can be accessed on a device where the key resides. It is recommended to use standard authentication flow to avoid risks associated with API Keys.',
```
alphadev4, line 8:
```suggestion
    more_info: 'API Keys should only be used for services in cases where other authentication methods are unavailable. Keys are insecure because they can be viewed publicly, such as from within a browser, or they can be accessed on a device where the key resides. It is recommended to use standard authentication flow to avoid risks associated with API Keys.',
```

alphadev4 (approved):

PR #1529, opened 2023-01-30
AWS Plugins for New CIS-Benchmarks

sm171190:
In this change we have added/modified AWS plugins as per CIS 1.5.0 benchamrks. We have added plugins  for 

- S3-BucketMFADelete Enabled and 
- EC2- NetworkACL Inboud Egress Rules. 
- We have expanded the audits for CloudWatch/MonitoringMetrics to include the Organizations Changes filter also.

### DIFF
--- a/plugins/aws/ec2/networkAclInboundTraffic.js
+++ b/plugins/aws/ec2/networkAclInboundTraffic.js
@@ -1,0 +1,90 @@
+var async = require('async');
+var helpers = require('../../../helpers/aws');
+
+module.exports = {
+    title: 'Unrestricted Network ACL Inbound Traffic',
+    category: 'EC2',
+    domain: 'Compute',
+    description: 'Ensures that no Amazon Network ACL allows inbound/ingress traffic to remote administration ports.',
+    more_info: 'Amazon Network ACL should not allow inbound/ingress traffic to remote administration ports to avoid unauthorized access at the subnet level.',
+    recommended_action: 'Update Network ACL to allow inbound/ingress traffic to specific port ranges only',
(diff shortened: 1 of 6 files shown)


plugins/aws/ec2/networkAclInboundTraffic.js
```diff
@@ -0,0 +1,90 @@
...
+        return true;
+    } else if (80>=range.From && 80<=range.To){ // HTTP
+        return true;
+    } else if (443>=range.From && 443<=range.To){ //HTTPS
+        return true;
+    } else if (53>=range.From && 53 <=range.To){ //DNS
+        return true;
+    }
+    return false;
+}
```
AkhtarAmir, line 90:
```suggestion
function disAllowedPorRange(range) {
    const allowedPorts = [22, 3389, 80, 443, 53];

    if (!range) return true;

    for (const port of allowedPorts) {
        if (port >= range.From && port <= range.To) return true;
    }

    return false;
}
```

AkhtarAmir (approved):

plugins/aws/s3/bucketMFADeleteEnabled.js
```diff
@@ -0,0 +1,82 @@
+var helpers = require('../../../helpers/aws');
+
+module.exports = {
+    title: 'S3 Bucket MFA Delete Status',
+    category: 'S3',
+    domain: 'Storage',
+    description: 'Ensures MFA delete is enabled on S3 buckets',
```
AkhtarAmir, line 7:
```suggestion
    description: 'Ensures MFA delete is enabled on S3 buckets.',
```

PR #1714, opened 2023-09-07
Map events to aws plugins

fatima99s:


plugins/aws/ec2/appTierInstanceIamRole.js
```diff
@@ -19,6 +19,7 @@ module.exports = {
             default: ''
         },
     },
+    realtime_triggers: ['ec2:RunInstance'],
```
sm171190, line 22: @AkhtarAmir Please note that this is ec2:RunInstances and not ec2:RunInstance. Let's change this and all other occurencs before merging

PR #1715, opened 2023-09-07
Update Dockerfile

jdosterman:
DOCKERFILE LINE 28
  ENTRYPOINT ["cloudsploitscan"] includes TYPO ... should be ["cloudsploit-scan"]


jdosterman:
It looks like the name of the link from /var/scan/node_modules/.bin has changed to back to cloudsploit-scan

PR #1770, opened 2023-11-30
SAAS-20297/ag-ssl-policy

alphadev4:


plugins/azure/applicationGateway/agSslPolicy.js
```diff
@@ -0,0 +1,54 @@
+const async = require('async');
+const helpers = require('../../../helpers/azure');
+
+module.exports = {
+    title: 'Application Gateway SSL Policy',
+    category: 'Application Gateway',
+    domain: 'Network Access Control',
+    description: 'Ensures that Application Gateway is using minimum TLS version of TLSv1_2',
+    more_info: 'Transport Layer Security (TLS), previously known as Secure Sockets Layer (SSL), is the standard security technology for establishing an encrypted link between a web server and a browser. This link ensures that all data passed between the web server and browsers remain private and encrypted.',
+    recommended_action: 'Modify Application Gateway with latest SSL policy which supports minimum TLS version',
```
alphadev4, line 10:
```suggestion
    recommended_action: 'Modify Application Gateway with latest SSL policy which supports minimum TLS version.',
```

plugins/azure/applicationGateway/agSslPolicy.spec.js
```diff
@@ -0,0 +1,101 @@
...
+                done();
+            });
+        });
+
+        it('should give passing result if Application Gateway is using ssl policy which supports minimum TLS version', function(done) {
+            const cache = createCache([appGateway[0]]);
+            agSslPolicy.run(cache, {}, (err, results) => {
+                expect(results.length).to.equal(1);
+                expect(results[0].status).to.equal(0);
+                expect(results[0].message).to.include('SSL policy  which supports TLSV1_2');
```
alphadev4, line 94:
```suggestion
                expect(results[0].message).to.include('SSL policy which supports TLSV1_2');
```

mehakseedat63 (changes requested):

plugins/azure/applicationGateway/agSslPolicy.js
```diff
@@ -0,0 +1,53 @@
...
+                helpers.addResult(results, 0, 'No existing Application Gateway found', location);
+                return rcb();
+            } 
+            
+            for (let appGateway of appGateways.data) {
+                if (!appGateway.id) continue;
+
+                const sslPolicy = appGateway.sslPolicy? appGateway.sslPolicy : '';
+                if ((sslPolicy.policyType == 'Predefined' && sslPolicy.policyName && recommendedSSLPolicies.indexOf(sslPolicy.policyName) > -1) ||
+                    (sslPolicy.policyType == 'Custom' && sslPolicy.minProtocolVersion && sslPolicy.minProtocolVersion.toLowerCase() == 'tlsv1_2')) {
```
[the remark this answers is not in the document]
alphadev4: but right in here we are checking the minimum tls version so should we add the check for tlsv1_3, if yes then should the message be like it is using latest TLS version?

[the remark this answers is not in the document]
mehakseedat63: Yes, this will fail in case of TLSV1_3, it should pass for v1_2 or higher. Update the messages accordingly

plugins/azure/applicationGateway/agSslPolicy.js
```diff
@@ -0,0 +1,73 @@
...
+                    if (found){
+                        helpers.addResult(results, 0, 'Application Gateway is using SSL policy which supports latest TLS version', location, appGateway.id);
+                    } else {
+                        helpers.addResult(results, 2, 'Application Gateway is using SSL policy which does not support latest TLS version', location, appGateway.id);
+                    }
+                } else {
+                    helpers.addResult(results, 2, 'Application Gateway is using SSL policy which does not support latest TLS version', location, appGateway.id);
+                } 
+            }
+
```
mehakseedat63, line 67:
```suggestion
                } else {
                    helpers.addResult(results, 2, 'Application Gateway is using SSL policy which does not support latest TLS version', location, appGateway.id);
                } 
            }

```
Duplicate code, maybe we can just set found to true or false inside the first if and then produce pass or fail results based off it outside the if?

mehakseedat63 (changes requested):

mehakseedat63 (approved):

PR #1780, opened 2023-12-01
ams classic v2 api disabled

abdullahaslam306:


plugins/azure/mediaServices/amsClassicApiDisabled.js
```diff
@@ -0,0 +1,59 @@
...
+                var getMediaService = helpers.addSource(cache, source, 
+                    ['mediaServices', 'get', location, mediaService.id]);
+
+                if (!getMediaService || getMediaService.err || !getMediaService.data) {
+                    helpers.addResult(results, 3, `Unable to query for Media Service: ${helpers.addError(getMediaService)}`,
+                        location, mediaService.id);
+                    continue;
+                }
+
+                if (getMediaService.data.identity) {
```
alphadev4, line 47: why are we checking identity here?

alphadev4, line 47: why are we checking identity here?

plugins/azure/mediaServices/amsClassicApiDisabled.js
```diff
@@ -0,0 +1,59 @@
...
+            }
+
+            for (let mediaService of mediaServices.data) {
+                if (!mediaService.id) continue;
+  
+                var getMediaService = helpers.addSource(cache, source, 
+                    ['mediaServices', 'get', location, mediaService.id]);
+
+                if (!getMediaService || getMediaService.err || !getMediaService.data) {
+                    helpers.addResult(results, 3, `Unable to query for Media Service: ${helpers.addError(getMediaService)}`,
```
alphadev4, line 42: the message for getMediaService here and listAll in case of unknown is same it should be Unable to query for Media Service data

plugins/azure/mediaServices/amsClassicApiDisabled.js
```diff
@@ -0,0 +1,59 @@
+var async = require('async');
+var helpers = require('../../../helpers/azure');
+
+module.exports = {
+    title: 'Media Services Classic API Disabled',
+    category: 'Media Services',
+    domain: 'Content Delivery',
+    description: 'Ensure that Microsoft Azure Media Services do not have the Classic API enabled.',
+    more_info: 'Disabling the Classic API for Azure Media Services is recommended to utilize modern APIs and features. Enabling classic features can enable the use of classic V2 APIs but might disable advanced security features like managed identities.',
+    link: 'https://learn.microsoft.com/en-us/azure/media-services/latest/migrate-v-2-v-3-differences-api-access',
```
alphadev4, line 10: i cannot find the keyword classic referenced with Media Service in this link? Can you add a link that defines what classic apis are and why we should not use them?

alphadev4 (changes requested):

plugins/azure/mediaServices/amsClassicApiDisabled.js
```diff
@@ -0,0 +1,59 @@
...
+                    ['mediaServices', 'get', location, mediaService.id]);
+
+                if (!getMediaService || getMediaService.err || !getMediaService.data) {
+                    helpers.addResult(results, 3, `Unable to query for Media Service: ${helpers.addError(getMediaService)}`,
+                        location, mediaService.id);
+                    continue;
+                }
+
+                if (getMediaService.data.identity) {
+                    helpers.addResult(results, 0, 'Media Service account has classic API disabled', location, mediaService.id);
```
mehakseedat63, line 48:
```suggestion
                    helpers.addResult(results, 0, 'Media Service account is not using classic v2 APIs', location, mediaService.id);
```

plugins/azure/mediaServices/amsClassicApiDisabled.js
```diff
@@ -0,0 +1,59 @@
...
+                } else {
+                    helpers.addResult(results, 2, 'Media Service account has classic API enabled', location, mediaService.id);
```
mehakseedat63, line 50:
```suggestion
                    helpers.addResult(results, 2, 'Media Service account is using classic v2 APIs', location, mediaService.id);
```

mehakseedat63 (changes requested):

plugins/azure/mediaServices/amsClassicApiDisabled.js
abdullahaslam306: This is only way to check if v2 apis are enabled or not because in case of v2 api we can not manage identity. Already discussed with Akhtar bhai

mehakseedat63 (approved):

alphadev4 (approved):

PR #1846, opened 2024-01-10
Adding support for gcp user account

v3rl:
Currently, Cloudsploit supports scanning using GCP service accounts.
These changes will add support to run Cloudsploit with GCP user account along with keeping with service account support intact.


v3rl:
To be honest the path of application_default_credentials.json provided in config_example.js is unnecessary but I have added as I did not want to break the LoadHelperfile function call. 
I changed JWT  to GoogleAuth object which grabs the required ADC credentials.

PR #1883, opened 2024-01-30
Blob service/cmk encryption

alphadev4:


plugins/azure/blobservice/blobContainersCmkEncrypted.js
```diff
@@ -0,0 +1,86 @@
...
+                        }).map(function(scope) {
+                            return scope.name;
+                        });
+                        var unencryptedCmkBlobs = blobContainers.data.filter(function(blob) {
+                            return !cmkEncryptionScopes.includes(blob.defaultEncryptionScope);
+                        }).map(function(blob){
+                            return blob.name;
+                        });
+                        if (unencryptedCmkBlobs && unencryptedCmkBlobs.length){
+                            helpers.addResult(results, 2, `Storage Account does not have CMK encryption enabled on following blob containers: ${unencryptedCmkBlobs.join(',')}`, location, storageAccount.id);
```
mehakseedat63, line 71:
```suggestion
                            helpers.addResult(results, 2, `Storage Account does not have CMK encryption enabled on following blob containers: ${unencryptedCmkBlobs.join(',')}`, location, storageAccount.id);
```
the resource should be the blob container itself and not the storage account

mehakseedat63 (changes requested):

plugins/azure/blobservice/blobContainersCmkEncrypted.js
```diff
@@ -0,0 +1,81 @@
...
+                            'Unable to query encryption scopes for Storage Accounts: ' + helpers.addError(encryptionScopes),
+                            location, storageAccount.id);
+                        break;
+                    } else {
+                        var cmkEncryptionScopes = encryptionScopes.data.filter(function(scope) {
+                            return scope.keyVaultProperties && scope.keyVaultProperties.keyUri;
+                        }).map(function(scope) {
+                            return scope.name;
+                        });
+                        blobContainers.data.filter(function(blob) {
```
mehakseedat63, line 64:
```suggestion
                        blobContainers.data.filter(function(blob) {
```

use for/foreach instead

mehakseedat63 (changes requested):

mehakseedat63 (approved):

PR #1886, opened 2024-01-30
plugin fixes

fatima99s:


plugins/azure/loadbalancer/lbHasTags.js
```diff
@@ -10,12 +10,26 @@ module.exports = {
     link: 'https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/tag-resources',
     recommended_action: 'Modify affected load balancers and add tags.',
     apis: ['loadBalancers:listAll'],
+    settings: {
+        ignore_internal_loadbalancers: {
```
mehakseedat63, line 14:
```suggestion
        ignore_internal_loadbalancers: {
```
setting name should be unique across all plugins

mehakseedat63 (changes requested):

mehakseedat63 (commented):
Sync with master and fix conflicts

plugins/azure/loadbalancer/lbHttpsOnly.js
```diff
@@ -15,13 +15,27 @@ module.exports = {
                 'App Service HTTPS redirection should be used to ensure site visitors ' +
                 'are always connecting over a secure channel.',
     },
+    settings: {
+        ignore_internal_loadbalancers: {
```
mehakseedat63, line 19:
```suggestion
        ignore_internal_lb_https: {
```
maybe we should add what we are checking for at the end instead of using multiple terms, name and description would stay the same. But attribute name will change.

mehakseedat63 (changes requested):

mehakseedat63 (approved):

PR #1910, opened 2024-02-18
Azure - OpenAI Account Managed Identity Enabled Plugin

mehakseedat63:


plugins/azure/openai/accountManagedIdentity.js
```diff
@@ -0,0 +1,55 @@
...
+            }
+
+            if (!accounts.data.length) {
+                helpers.addResult(results, 0, 'No existing OpenAI accounts found', location);
+                return rcb();
+            }
+
+            for (let account of accounts.data) {
+
+                if (account.identity && account.identity.type &&
```
alphadev4, line 39: it looks good but i think its fine without adding the checks of identity types like we have in other plugins

mehakseedat63: if we enable then disable it then it returns type: "none"
Screenshot 2024-02-19 at 3 35 47 PM

alphadev4 (approved):

PR #1945, opened 2024-03-02
Azure/AG-https-listeners

alphadev4:


plugins/azure/applicationGateway/agHttpsListenerOnly.js
```diff
@@ -0,0 +1,53 @@
+const async = require('async');
+const helpers = require('../../../helpers/azure');
+
+module.exports = {
+    title: 'Application Gateway HTTPS Listener',
+    category: 'Application Gateway',
+    domain: 'Network Access Control',
+    severity: 'Medium',
+    description: 'Ensures that Application Gateway is not configured to use any non-HTTPS listener.',
```
mehakseedat63, line 9:
```suggestion
    description: 'Ensures that Application Gateway is configured to use only HTTPS listeners.',
```
I think phrasing it this way adds more clarity

plugins/azure/applicationGateway/agHttpsListenerOnly.js
```diff
@@ -0,0 +1,53 @@
...
+                return rcb();
+            }
+
+            if (!appGateways.data.length) {
+                helpers.addResult(results, 0, 'No existing Application Gateway found', location);
+                return rcb();
+            }
+
+            for (let appGateway of appGateways.data) {
+                if (!appGateway.id) continue;
```
mehakseedat63, line 38:
```suggestion
                if (!appGateway.id) continue;
                
```

plugins/azure/applicationGateway/agHttpsListenerOnly.js
```diff
@@ -0,0 +1,53 @@
...
+                var httpListeners = appGateway.httpListeners ? appGateway.httpListeners.filter(listener => listener.protocol && listener.protocol.toLowerCase() != 'https')
```
mehakseedat63, line 39:
```suggestion
                if (appGateway.httpListeners && appGateway.httpListeners.length) {
                     var httpListeners = appGateway.httpListeners.filter(listener => listener.protocol && listener.protocol.toLowerCase() != 'https').map(listener => listener.name)
                }
               
```
if we're passing in case of no listeners then maybe we should handle this check like this

plugins/azure/applicationGateway/agHttpsListenerOnly.js
```diff
@@ -0,0 +1,53 @@
...
+                    .map(listener => listener.name) : [];
+                if (httpListeners && httpListeners.length) {
+                    helpers.addResult(results, 2, `Application Gateway is using following non-https listeners: ${httpListeners.join(',')}`, location, appGateway.id);
+                } else {
+                    helpers.addResult(results, 0, 'Application Gateway is not using any non-https listener', location, appGateway.id);
```
mehakseedat63, line 44:
```suggestion
                    helpers.addResult(results, 0, 'Application Gateway is using https listeners only', location, appGateway.id);
```

mehakseedat63 (changes requested):

PR #2108, opened 2024-11-06
Revised keyVaultSecretExpiry

AkhtarAmir:


plugins/azure/keyvaults/keyVaultSecretExpiry.js
```diff
@@ -46,6 +46,13 @@ module.exports = {
             }
 
             vaults.data.forEach(function(vault) {
+                // Check if vault is RBAC-enabled
+                if (!vault.properties || !vault.properties.enableRbacAuthorization) {
+                    helpers.addResult(results, 0,
```
fatima99s, line 51: can we skip the results for non rbac key?

alphadev4 (approved):

PR #2114, opened 2024-11-08
TLS Version change

AkhtarAmir:


plugins/azure/storageaccounts/storageAccountsTlsVersion.js
```diff
@@ -33,7 +25,7 @@ module.exports = {
         var locations = helpers.locations(settings.govcloud);
 
         var config = {
```
fatima99s, line 27: can we assign like 
 var tlsVersion = 1.2
 instead of putting it in config object

plugins/azure/eventhub/eventHubMinimumTLSversion.js
```diff
@@ -27,7 +19,7 @@ module.exports = {
         var locations = helpers.locations(settings.govcloud);
 
         var config = {
```
fatima99s, line 21:
```suggestion

```

PR #2125, opened 2025-02-03
added validation for suppress regex

AkhtarAmir:

PR #2132, opened 2025-03-07
fix: allow to use wildcards in pluginId again

### LINKED ISSUE #2127: Wildcard cannot be used in place of pluginId in suppression argument after recent changes
After changes made in https://github.com/aquasecurity/cloudsploit/pull/2125 the next command (which also can be found in `README.md`) cannot be run as it's not possible to use wildcard in place of `pluginId` anymore:

```sh
# Suppress all us-east-1 region results
$ ./index.js --suppress *:us-east-1:*
```

Returns:

```sh
Error: Invalid pluginId in suppression: *. Must only contain letters and numbers and be between 1-255 characters.
```

pbilyak:
This PR restores the ability to use wildcards in `pluginId`, which was removed in #2125.

fixes #2127

### DIFF
--- a/postprocess/suppress.js
+++ b/postprocess/suppress.js
@@ -15,7 +15,7 @@ module.exports = {
                 throw new Error(`Invalid suppression format: ${expr}. Expected format: pluginId:region:resourceId`);
             }
             
-            const pluginPattern = /^[A-Za-z0-9]{1,255}$/; // eslint-disable-line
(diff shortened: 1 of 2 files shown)


pbilyak:
fixed in https://github.com/aquasecurity/cloudsploit/commit/c8abde966b0734e635660460aec9e86b1e0ccbe2

PR #2188, opened 2026-02-16
Fix Azure Storage Account Public Network Access plugin

Taha-Aaqib:
Fix storage account public network access check to detect open CIDR ranges and check if selected networks are given access.


alphadev4 (approved):

helpers/azure/resources.js
```diff
@@ -269,7 +269,7 @@ module.exports = {
         list: 'id'
     },
     serviceBus:{
-        listNamespacesBySubscription: 'id'
+        listNamespacesBySubscription: 'id',
```
alphadev4:
```suggestion
        listNamespacesBySubscription: 'id'
```

alphadev4 (changes requested):

PR #2205, opened 2026-08-31
SLK-133892/azure-plugins-draft-1

Hamza-Aziz61:


helpers/azure/api.js
```diff
@@ -510,10 +517,29 @@ var calls = {
     },
     users: {
         list: {
-            url: 'https://graph.microsoft.com/v1.0/users',
+            url: 'https://graph.microsoft.com/v1.0/users?$select=id,displayName,userPrincipalName,mail,userType,accountEnabled',
```
Meerab-Shafique, line 520: Did you test the existing plugin after this change?

Meerab-Shafique: Ensure No Guest User

plugins/azure/databricks/workspacePublicAccess.js
```diff
@@ -0,0 +1,53 @@
...
+var helpers = require('../../../helpers/azure');
+
+module.exports = {
+    title: 'Databricks Workspace Public Access',
+    category: 'AI & ML',
+    domain: 'Machine Learning',
+    severity: 'Medium',
+    description: 'Ensures that Azure Databricks Workspace has public network access disabled.',
+    more_info: 'Disabling public network access ensures that the Databricks workspace is not reachable over the public internet and can only be accessed through private endpoints within trusted networks. This reduces the attack surface and the risk of unauthorized access.',
+    recommended_action: 'Modify Databricks workspace networking settings and set Allow Public Network Access to Disabled.',
```
Meerab-Shafique:
```suggestion
    recommended_action: 'Modify Databricks workspace networking settings and set Allow Public Network Access to disabled.',
```

helpers/azure/api.js
Hamza-Aziz61: yes, I tested both the existing plugins and the new one

plugins/azure/databricks/workspacePublicAccess.js
```diff
@@ -0,0 +1,53 @@
...
+    link: 'https://learn.microsoft.com/en-us/azure/databricks/security/network/front-end/front-end-private-connect',
```
Meerab-Shafique, line 12: do you think this endpoint is more valid? https://learn.microsoft.com/en-us/azure/databricks/security/network/classic/secure-cluster-connectivity ?

plugins/azure/entraid/subscriptionOwnerCount.js
```diff
@@ -0,0 +1,88 @@
+const async = require('async');
+const helpers = require('../../../helpers/azure');
+
+module.exports = {
+    title: 'Subscription Owner Count',
+    category: 'Entra ID',
+    domain: 'Identity and Access Management',
+    severity: 'Medium',
```
Meerab-Shafique:
```suggestion
    severity: 'Low',
```

plugins/azure/entraid/subscriptionOwnerCount.js
```diff
@@ -0,0 +1,88 @@
...
+    more_info: 'The Owner role grants full control over all resources in a subscription, including the ability to assign roles to others. Keeping the number of owners low limits privilege sprawl, while keeping more than one avoids losing administrative access. All principal types count towards the total, including users, groups, service principals and managed identities.',
+    recommended_action: 'Remove unnecessary Owner role assignments, or add an additional owner if only one exists.',
+    link: 'https://learn.microsoft.com/en-us/azure/role-based-access-control/built-in-roles#owner',
+    apis: ['roleDefinitions:list', 'aad:listRoleAssignments'],
+    settings: {
+        subscription_owners_min: {
+            name: 'Subscription Owners Minimum',
+            description: 'Return a failing result when the number of subscription owners is below this value',
+            regex: '^[1-9]{1}[0-9]{0,3}$',
+            default: 2
```
Meerab-Shafique, line 19:
```suggestion
            default: 1
```

Meerab-Shafique: the minimum number should be 1

plugins/azure/entraid/subscriptionOwnerCount.js
```diff
@@ -0,0 +1,88 @@
...
+        },
+        subscription_owners_max: {
+            name: 'Subscription Owners Maximum',
+            description: 'Return a failing result when the number of subscription owners exceeds this value',
+            regex: '^[1-9]{1}[0-9]{0,3}$',
+            default: 3
```
Meerab-Shafique, line 25: the default for maximum can be 2 and if the customers want they can increase it

plugins/azure/entraid/userAccessAdminRestricted.js
```diff
@@ -0,0 +1,69 @@
+const async = require('async');
+const helpers = require('../../../helpers/azure');
+
+module.exports = {
+    title: 'User Access Administrator Role Restricted',
+    category: 'Entra ID',
+    domain: 'Identity and Access Management',
+    severity: 'Medium',
```
Meerab-Shafique, line 8:
```suggestion
    severity: 'Low',
```

should this be low ? or medium

plugins/azure/entraid/userAccessAdminRestricted.js
```diff
@@ -0,0 +1,69 @@
...
+            var adminAssignments = roleAssignments.data.filter(roleAssignment => roleAssignment.roleDefinitionId &&
+                adminRoleIds.includes(roleAssignment.roleDefinitionId.split('/').pop()));
+
+            if (!adminAssignments.length) {
+                helpers.addResult(results, 0, 'User Access Administrator role is not assigned', location);
+                return rcb();
+            }
+
+            adminAssignments.forEach(roleAssignment => {
+                helpers.addResult(results, 2, 'User Access Administrator role is assigned', location, roleAssignment.id);
```
Meerab-Shafique: what does this plugin change? at which level? it means that this role can never be assigned?

plugins/azure/storageaccounts/fileShareSmbChannelEncryption.js
```diff
@@ -0,0 +1,71 @@
+const async = require('async');
+const helpers = require('../../../helpers/azure');
+
+module.exports = {
+    title: 'File Share SMB Channel Encryption',
+    category: 'Storage Accounts',
+    domain: 'Storage',
+    severity: 'Medium',
+    description: 'Ensures that SMB file shares are configured to only allow AES-256-GCM or higher for SMB channel encryption.',
```
Meerab-Shafique, line 9: how do we know which encryptions are higher?

plugins/azure/storageaccounts/fileShareSmbChannelEncryption.js
```diff
@@ -0,0 +1,71 @@
...
+                        location, storageAccount.id);
+                } else {
+                    const channelEncryption = getServiceProperties.data.protocolSettings &&
+                        getServiceProperties.data.protocolSettings.smb &&
+                        getServiceProperties.data.protocolSettings.smb.channelEncryption;
+
+                    const encryptionList = channelEncryption ?
+                        channelEncryption.split(';').map(algorithm => algorithm.trim()).filter(algorithm => algorithm) : [];
+
+                    if (encryptionList.length && encryptionList.every(algorithm => algorithm.toUpperCase() === 'AES-256-GCM')) {
```
Meerab-Shafique, line 54: this will only pass if the encryption channel is this, not higher?

plugins/azure/storageaccounts/fileShareSmbProtocolVersion.js
```diff
@@ -0,0 +1,71 @@
+const async = require('async');
+const helpers = require('../../../helpers/azure');
+
+module.exports = {
+    title: 'File Share SMB Protocol Version',
+    category: 'Storage Accounts',
+    domain: 'Storage',
+    severity: 'Medium',
+    description: 'Ensures that SMB file shares are configured to only allow SMB 3.1.1 or higher.',
```
Meerab-Shafique, line 9: Same concern how do we check higher

plugins/azure/storageaccounts/geoRedundantStorage.js
```diff
@@ -0,0 +1,65 @@
...
+            }
+
+            storageAccounts.data.forEach(storageAccount => {
+                if (!storageAccount.id) return;
+
+                const skuName = storageAccount.sku ? storageAccount.sku.name : undefined;
+
+                if (!skuName) {
+                    helpers.addResult(results, 3,
+                        'Unable to determine Storage Account redundancy setting',
```
Meerab-Shafique, line 47: sku tells us the storage account redundancy settings? i think it tells us about the tier of storage account

plugins/azure/storageaccounts/storageAccountEntraIdAuthDefault.js
```diff
@@ -0,0 +1,57 @@
+const async = require('async');
+const helpers = require('../../../helpers/azure');
+
+module.exports = {
+    title: 'Storage Account Entra ID Authorization Default',
```
Meerab-Shafique, line 5: is this different than Storage Accounts Entra ID Enabled?

Meerab-Shafique (changes requested):

plugins/azure/databricks/workspacePublicAccess.js
Hamza-Aziz61: your endpoint is more about enableNoPublicIp which is different from Public Access which we are targeting in this plugin, so i dont think we should change it

plugins/azure/entraid/subscriptionOwnerCount.js
Hamza-Aziz61: but CIS explicitly states that "5.7 Ensure there are between 2 and 3 Subscription Owners", so I don't think we should change it but I can if you say so

plugins/azure/entraid/subscriptionOwnerCount.js
Hamza-Aziz61: same as above cis says between 2 and 3

plugins/azure/entraid/userAccessAdminRestricted.js
Hamza-Aziz61: according to CIS it says "Unnecessary assignments heighten the risk of privilege escalation and unauthorized access",
so I think it should remain medium

plugins/azure/entraid/userAccessAdminRestricted.js
Hamza-Aziz61: the plugin checks that whether the User Access Administrator role is assigned or not, but it is checking at any level I have updated the plugin to only check at root level as CIS says in its control and it doesn't mean it can never be assigned it means it should not remain after its use, it should be removed

plugins/azure/storageaccounts/fileShareSmbChannelEncryption.js
Hamza-Aziz61: CIS says "9.1.3 Ensure 'SMB channel encryption' is Set to 'AES-256-GCM' or Higher for SMB file shares" , but the max encryption is AES-256-GCM nothing higher is available CIS basically future proofed the title and i used same wording, should i change it to only AES-256-GCM?

plugins/azure/storageaccounts/fileShareSmbChannelEncryption.js
Hamza-Aziz61: as there is no higher encryption i hardcoded AES-256-GCM, we don't know the name of higher encryption that will be released so we will have to update the plugin when it is released

plugins/azure/storageaccounts/fileShareSmbProtocolVersion.js
Hamza-Aziz61: same as above 3.1.1 is the highest version in azure we don't know what higher version will look like so i hardcoded it, i can change the description if you like

plugins/azure/storageaccounts/geoRedundantStorage.js
Hamza-Aziz61: sku.tier tells the tier but this plugin is reading sku.name which gives redundancy settings

plugins/azure/storageaccounts/storageAccountEntraIdAuthDefault.js
Hamza-Aziz61: yes this plugin Storage Accounts Entra ID Enabled reads enableAzureFilesAadIntegration that is about mounting Azure Files (SMB) using AD/Entra Domain Services but the pr plugin reads defaultToOAuthAuthentication to check whether the Azure Portal UI defaults to Entra ID

plugins/azure/storageaccounts/fileShareSmbChannelEncryption.js
Meerab-Shafique: lets add/maintain the array for now for encryption and in future if there is a higher encryption we can add it in the same array

plugins/azure/storageaccounts/geoRedundantStorage.js
Meerab-Shafique: how will sku.name will give us redundancy setting?

Hamza-Aziz61: this is from official azure doc it gives what we want
image

fatimaaqua (dismissed):

plugins/azure/storageaccounts/fileShareSmbChannelEncryption.js
Hamza-Aziz61: done

Meerab-Shafique (approved):

